PR #24, opened 2020-10-30
fix for dynatrace manged urls

pawelsiwek:
Fix to support URL's in Dynatrace managed format: https://abcd.managed-dt.dynalabs.io/e/ett29184-1ae2-46dd-96fe-02957773fc57/ 
Using urljoin removes all URL segments .. :(

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -112,12 +112,12 @@ s)
 esac
 
 echo "Please provide the URL used to access Dynatrace, for example: https://mytenant.live.dynatrace.com/"
-while ! [[ "${DYNATRACE_URL}" =~ ^https?:\/\/[-a-zA-Z0-9@:%._+~=]{1,256}\/$ ]]; do
+while ! [[ "${DYNATRACE_URL}" =~ ^(https?:\/\/[-a-zA-Z0-9@:%._+~=]{1,256}\/)(e\/[a-z0-9-]{36}\/)?$ ]]; do
     read -p "Enter Dynatrace tenant URI: " DYNATRACE_URL
 done
 echo ""
 
-echo "Please log in to Dynatrace, and generate API token (Settings->Integration->Dynatrace API). The token requires grant of 'API v2 Ingest metrics', 'API v1 Read configuration' and `WriteConfig` scope"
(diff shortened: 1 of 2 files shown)


sq2gxo (approved):

scripts/setup.sh
```diff
@@ -112,12 +112,12 @@ s)
...
 
 echo "Please provide the URL used to access Dynatrace, for example: https://mytenant.live.dynatrace.com/"
-while ! [[ "${DYNATRACE_URL}" =~ ^https?:\/\/[-a-zA-Z0-9@:%._+~=]{1,256}\/$ ]]; do
+while ! [[ "${DYNATRACE_URL}" =~ ^(https?:\/\/[-a-zA-Z0-9@:%._+~=]{1,256}\/)(e\/[a-z0-9-]{36}\/)?$ ]]; do
     read -p "Enter Dynatrace tenant URI: " DYNATRACE_URL
 done
 echo ""
 
-echo "Please log in to Dynatrace, and generate API token (Settings->Integration->Dynatrace API). The token requires grant of 'API v2 Ingest metrics', 'API v1 Read configuration' and `WriteConfig` scope"
+echo "Please log in to Dynatrace, and generate API token (Settings->Integration->Dynatrace API). The token requires grant of 'API v2 Ingest metrics', 'API v1 Read configuration' and 'WriteConfig' scope"
```
sq2gxo, line 120: "Read configuration' and 'WriteConfig' scope" -> "'Read configuration' and 'Write configuration' scope"?

src/lib/metric_ingest.py
```diff
@@ -59,7 +59,7 @@ async def _push_to_dynatrace(context: Context, lines_batch: List[IngestLine]):
         context.log("Ingest input is: ")
         context.log(ingest_input)
     ingest_response = await context.session.post(
-        url=urljoin(context.dynatrace_url, "/api/v2/metrics/ingest"),
+        url=f"{context.dynatrace_url}/api/v2/metrics/ingest",
```
pgutkowski, line 62: If we want to do a simple string appending, then we need to handle double slash case - if `context.dynatrace_url` ends with `/`, then result path has empty URL segment and Dynatrace server returns 404.

Something like that should work
```suggestion
        url=f"{context.dynatrace_url.rstrip('/')}/api/v2/metrics/ingest",
```

pgutkowski (changes requested):

pgutkowski (approved):

PR #34, opened 2020-12-02
Implement GCP FastCheck upon container start

r-orzechowski:


src/lib/fast_check.py
```diff
@@ -0,0 +1,123 @@
...
+                return None
+
+            return await response.json()
+        except Exception as e:
+            self.logging_context.log(f'Unable to get Dynatrace token metadata. Error details: {e}')
+            return None
+
+    async def _check_services(self, project_id):
+        list_services_result = await self.list_services(project_id)
+        service_names = [find_service_name(service['name']) for service in list_services_result['services']] if list_services_result else []
```
pgutkowski, line 81: `list_services_result` could return empty list instead of `None` when no services are detected, so no null check would be required

src/lib/fast_check.py
```diff
@@ -0,0 +1,123 @@
...
+        if not all(name in service_names for name in required_services):
+            self.logging_context.log(f'Cannot monitor project: \'{project_id}\'. '
+                                     f'Enable required services: {required_services}')
+            return None
+        return service_names
+
+    async def _check_dynatrace(self, project_id):
+        try:
+            dynatrace_url = await fetch_secret(self.session, project_id, self.token, "DYNATRACE_URL")
```
pgutkowski, line 90: You should use `fetch_dynatrace_url` and `fetch_dynatrace_api_key` functions from `credentials.py` instead. They fetch the relevant secret name from environment variables.

src/lib/fast_check.py
```diff
@@ -0,0 +1,123 @@
...
+            dynatrace_access_key = await fetch_secret(self.session, project_id, self.token, "DYNATRACE_ACCESS_KEY")
+            if not dynatrace_url or not dynatrace_access_key:
+                self.logging_context.log(f'No Dynatrace secrets: DYNATRACE_URL, DYNATRACE_ACCESS_KEY for project: {project_id}.'
+                                         f'Add required secrets to Secret Manager.')
+                return None
+
+            token_metadata = await self.get_dynatrace_token_metadata(dynatrace_url, dynatrace_access_key)
+            if token_metadata['revoked'] or not all(scope in token_metadata['scopes'] for scope in dynatrace_required_token_scopes):
```
pgutkowski, line 98: You could extract `all(scope in token_metadata['scopes'] for scope in dynatrace_required_token_scopes)` into function with descriptive name

src/run_docker.py
```diff
@@ -13,19 +13,42 @@
...
 
-async def scheduling_loop():
+async def scheduling_loop(project_ids: Optional[List[str]] = None):
     while True:
-        loop.create_task(async_dynatrace_gcp_extension())
+        loop.create_task(async_dynatrace_gcp_extension(project_ids))
         await asyncio.sleep(60)
 
 
+async def init():
```
pgutkowski, line 34: name `init` is rather nebulous.

src/lib/fast_check.py
mswiatkowska, line 81: Minor: you can use if statement in list comprehension (and skip else part then)
[find_service_name(service['name']) for service in list_services_result['services'] if list_services_result]

pgutkowski (changes requested):

src/lib/fast_check.py
r-orzechowski: Improved code here; got rid of cumbersome if check and return empty list; have a look.

r-orzechowski: Similar comment form @pgutkowski. Have a look at the improvement there.

pgutkowski:
@r-orzechowski you probably should rebase your changes on `master`, because of the merge you've made it looks like your PR has 17 commits, which already are in `master`.

pgutkowski (approved):

mswiatkowska (approved):

pgutkowski (approved):

PR #38, opened 2020-12-09
APM-272389 fix RATE_LIMIT_EXCEEDED problem

wojtasskorcz:
Please check if the implementation is sound. If so, I'll add documentation and tests (if possible)

### DIFF
--- a/src/lib/context.py
+++ b/src/lib/context.py
@@ -78,6 +78,7 @@ class Context(LoggingContext):
         self.maximum_metric_data_points_per_minute = int(os.environ.get("MAXIMUM_METRIC_DATA_POINTS_PER_MINUTE", "100000"))
         self.metric_ingest_batch_size = int(os.environ.get("METRIC_INGEST_BATCH_SIZE", "1000"))
         self.require_valid_certificate = os.environ.get("REQUIRE_VALID_CERTIFICATE", "True") in ["True", "T", "true"]
+        self.use_x_goog_user_project_header = {}
 
         # self monitoring data
         self.dynatrace_request_count = {}
--- a/src/lib/context.py
+++ b/src/lib/context.py
@@ -78,7 +78,7 @@ class Context(LoggingContext):
         self.maximum_metric_data_points_per_minute = int(os.environ.get("MAXIMUM_METRIC_DATA_POINTS_PER_MINUTE", "100000"))
         self.metric_ingest_batch_size = int(os.environ.get("METRIC_INGEST_BATCH_SIZE", "1000"))
         self.require_valid_certificate = os.environ.get("REQUIRE_VALID_CERTIFICATE", "True") in ["True", "T", "true"]
-        self.use_x_goog_user_project_header = {}
+        self.use_x_goog_user_project_header = {project_id_owner: False}
 
         # self monitoring data
         self.dynatrace_request_count = {}
--- a/src/lib/metric_ingest.py
+++ b/src/lib/metric_ingest.py
@@ -145,6 +145,8 @@ async def fetch_metric(
     headers = {
         "Authorization": "Bearer {token}".format(token=context.token)
     }
+    if context.use_x_goog_user_project_header[project_id]:
+        headers["x-goog-user-project"] = project_id
 
     should_fetch = True
 
--- a/src/lib/metric_ingest.py
+++ b/src/lib/metric_ingest.py
@@ -147,7 +147,7 @@ async def fetch_metric(
     headers = {
         "Authorization": "Bearer {token}".format(token=context.token)
     }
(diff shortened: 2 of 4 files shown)


src/main.py
```diff
@@ -138,13 +138,41 @@ async def handle_event(event: Dict, event_context, project_id_owner: Optional[st
...
+    }
+    resp = await context.session.get(url=url, params=params, headers=headers)
+    page = await resp.json()
+
+    if resp.status == 200:
+        context.use_x_goog_user_project_header[project_id] = True
+    elif resp.status == 403 and 'serviceusage.services.use' in page['error']['message']:
+        context.use_x_goog_user_project_header[project_id] = False
+    else:
+        raise Exception(str(page))
```
pgutkowski, line 173: We definitely don't want to raise any exceptions, that would prevent any polling. Please replace with some log message.

src/main.py
```diff
@@ -138,13 +138,41 @@ async def handle_event(event: Dict, event_context, project_id_owner: Optional[st
 
 async def process_project_metrics(context: Context, project_id: str, services: List[GCPService]):
     context.log(project_id, f"Starting processing...")
+    await check_x_goog_user_project_header_permissions(context, project_id)
```
pgutkowski, line 141: Please wrap with try - we don't want this check to prevent the polling.

src/main.py
```diff
@@ -138,13 +138,41 @@ async def handle_event(event: Dict, event_context, project_id_owner: Optional[st
...
 
+async def check_x_goog_user_project_header_permissions(context: Context, project_id: str):
+    if project_id in context.use_x_goog_user_project_header:
+        return
+
+    service_usage_booking = os.environ['SERVICE_USAGE_BOOKING'] if 'SERVICE_USAGE_BOOKING' in os.environ.keys() \
+        else 'source'
+    if service_usage_booking != 'destination':
+        context.use_x_goog_user_project_header[project_id] = False
+        return
```
pgutkowski, line 157: Maybe I misunderstood, but I thought that we want to perform the check by default, and have an option to override it with env variable to avoid API calls. Now SERVICE_USAGE_BOOKING == "destination" is required to perform check at all.
@pawelsiwek what's you opinion?

src/lib/metric_ingest.py
```diff
@@ -147,6 +147,8 @@ async def fetch_metric(
     headers = {
         "Authorization": "Bearer {token}".format(token=context.token)
     }
+    if context.use_x_goog_user_project_header[project_id]:
```
pgutkowski, line 150: please use `get` with default to avoid KeyError

src/main.py
```diff
@@ -138,13 +138,41 @@ async def handle_event(event: Dict, event_context, project_id_owner: Optional[st
...
     context.fetch_gcp_data_execution_time[project_id] = fetch_data_time
     context.log(project_id, f"Finished fetching data in {fetch_data_time}")
     await push_ingest_lines(context, project_id, ingest_lines)
 
 
+async def check_x_goog_user_project_header_permissions(context: Context, project_id: str):
+    if project_id in context.use_x_goog_user_project_header:
+        return
+
+    service_usage_booking = os.environ['SERVICE_USAGE_BOOKING'] if 'SERVICE_USAGE_BOOKING' in os.environ.keys() \
```
pgutkowski, line 153: Please add `SERVICE_USAGE_BOOKING` env variable to table in HACKING.MD. @pawelsiwek do we want to update the README with info about new permission as well?

pgutkowski (changes requested):

src/main.py
wojtasskorcz: I applied fixes requested in your other comments. As I understood it, the following must be true to enable the per-project billing:
- `SERVICE_USAGE_BOOKING` env var is set to `destination`
- SA must have permission to set the `x-goog-user-project` header
Otherwise nothing changes in billing.
Happy to change this logic when clarified by @pawelsiwek

src/lib/context.py
```diff
@@ -78,6 +78,7 @@ def __init__(
         self.maximum_metric_data_points_per_minute = int(os.environ.get("MAXIMUM_METRIC_DATA_POINTS_PER_MINUTE", "100000"))
         self.metric_ingest_batch_size = int(os.environ.get("METRIC_INGEST_BATCH_SIZE", "1000"))
         self.require_valid_certificate = os.environ.get("REQUIRE_VALID_CERTIFICATE", "True") in ["True", "T", "true"]
+        self.use_x_goog_user_project_header = {}
```
pgutkowski, line 81: There is no point in sending this header to owner project, we can optimize a bit by initializing this Dict with `{project_id_owner: False}`

pgutkowski (approved):

PR #40, opened 2020-12-18
import dashboard for k8s + owner tag required for v208

pawelsiwek:


src/run_docker.py
```diff
@@ -48,6 +49,11 @@ async def initial_check():
             logging_context.log(f'Monitoring disabled. Unable to acquire authorization token.')
     await session.close()
 
+async def try_configure_dynatrace():
+    async with aiohttp.ClientSession() as session:
+        dashboards_result = await ConfigureDynatrace(session, logging_context)
+    await session.close()
```
pgutkowski, line 55: this `close()` call is superfluous when you are already using `with` in line 53

src/lib/configure_dynatrace.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def __init__(self, session: ClientSession, logging_context: LoggingContext):
+        self.session = session
+        self.logging_context = logging_context
+            
+    async def _init_(self):        
+        dynatrace_url = await fetch_dynatrace_url(self.session, "", "")
+        dynatrace_access_key = await fetch_dynatrace_api_key(self.session,"", "")
+        scopes = await  get_dynatrace_token_metadata(self.session, self.logging_context, dynatrace_url, dynatrace_access_key)
+        has_write_config_permission=any(s in scopes.get('scopes') for s in ["WriteConfig", "ReadConfig"])
```
pgutkowski, line 107: Please use `scopes.get('scopes', [])` to avoid KeyError if `get_dynatrace_token_metadata` returns empty dict on request exception

src/lib/configure_dynatrace.py
```diff
@@ -0,0 +1,123 @@
...
+        has_write_config_permission=any(s in scopes.get('scopes') for s in ["WriteConfig", "ReadConfig"])        
+        if not has_write_config_permission:
+            self.logging_context.log("Missing ReadConfig/WriteConfig permission for Dynatrace API token, skipping dashboards configuration")
+        else:
+            existing_dashboards = await self.get_existing_dashboards(dynatrace_url, dynatrace_access_key)                                
+            available_dashboards = await self.get_available_dashboards()
+            dashboards_to_install = [dash for dash in available_dashboards if dash["name"] not in existing_dashboards]
+
+            self.logging_context.log(f"Available dashboards: {[dash['name'] for dash in available_dashboards]}")
+            if len(dashboards_to_install) > 0 :
```
pgutkowski, line 116: `if dashboards:` would be more concise

src/lib/configure_dynatrace.py
```diff
@@ -0,0 +1,123 @@
...
+                self.logging_context.log(f'Unable to create dashboard {name} in Dynatrace: {response.status}, url: {response.url}, reason: {response.reason}, message {response_json}')                                
+                return {}
+            else:
+                self.logging_context.log(f"Installed dashboard {name}")
+            return {}
+        except Exception as e:
+            self.logging_context.log(f'Unable to create dashboard in Dynatrace. Error details: {e}')
+            return {}
+
+    async def get_existing_dashboards(self, dynatrace_url: str, dynatrace_api_key: str, timeout: Optional[int] = 2) -> dict:
```
pgutkowski, line 44: Please change return value type to `List` and fix lines 55 and 62 - we are returning different type on successful (List of dicts) and failed (empty dict) call.

src/lib/configure_dynatrace.py
```diff
@@ -0,0 +1,123 @@
...
+
+            dashboards_json=  await response.json()
+            all_dashboards = [dashboard["name"] for dashboard in dashboards_json.get("dashboards",[])]
+            return all_dashboards
+        except Exception as e:
+            self.logging_context.log(f'Unable to get existing dashboards config. Error details: {e}')
+            return {}
+
+
+    async def get_available_dashboards(self) -> dict:
```
pgutkowski, line 65: Please change return value type to List and fix line 96 - we are returning different type on successful (List of dicts) and failed (empty dict) call.

src/lib/configure_dynatrace.py
```diff
@@ -0,0 +1,123 @@
...
+        try:
+            selected_services = None
+            if "GCP_SERVICES" in os.environ:
```
pgutkowski, line 68: Correct me if I'm wrong, but it seems that we are not returning at all, if this check is negative. That would make code in 113 fail. Please `add return []` line at the bottom of this function body.

src/lib/configure_dynatrace.py
```diff
@@ -0,0 +1,123 @@
...
+                },
+                timeout=timeout,
+                json=dashboard_json)
+            if response.status != 201:
+                response_json=  await response.json()
+                self.logging_context.log(f'Unable to create dashboard {name} in Dynatrace: {response.status}, url: {response.url}, reason: {response.reason}, message {response_json}')                                
+                return {}
+            else:
+                self.logging_context.log(f"Installed dashboard {name}")
+            return {}
```
pgutkowski, line 39: Why are we returning any value in this function, if it is always the empty dict?

pgutkowski (changes requested):

pgutkowski (approved):

src/lib/configure_dynatrace.py
```diff
@@ -0,0 +1,136 @@
...
+                    "Authorization": f"Api-Token {dynatrace_api_key}",
+                    "Content-Type": "application/json; charset=utf-8"
+                },
+                timeout=timeout,
+                json=dashboard_json)
+            if response.status != 201:
+                response_json=  await response.json()                
+                if 'owner' in json.dumps(response_json):                                        
+                    del dashboard_json["dashboardMetadata"]["owner"]
+                    response = await self.session.post(
```
pgutkowski, line 36: To make this more readable you could add `remove_owner: bool = False` argument to `post_dashboard` function. Function could check this argument to determine whether to perform ` del dashboard_json["dashboardMetadata"]["owner"]` before sending request, and to check if it should retry without `owner` field.

pgutkowski (changes requested):

src/lib/configure_dynatrace.py
```diff
@@ -0,0 +1,126 @@
...
+                url=f"{dynatrace_url.rstrip('/')}/api/config/v1/dashboards",
+                headers={
+                    "Authorization": f"Api-Token {dynatrace_api_key}",
+                    "Content-Type": "application/json; charset=utf-8"
+                },
+                timeout=timeout,
+                json=dashboard_json)
+            if response.status != 201:
+                response_json=  await response.json()                
+                if 'owner' in json.dumps(response_json):
```
pgutkowski, line 36: Please change to `if not remove_owner and "owner" in json.dumps(response_json)` to avoid falling into recurrent loop.

pgutkowski (approved):

PR #46, opened 2021-02-12
All services preview

pawelsiwek:
Added featureSet concept to solve problem with gce_instance / k8s_container lots of metrics , giving ability to opt-in for detailed ones and have default limited set enabled


mfranczak:
@pawelsiwek I think we are missing documentation on how to extend the amount of services we support. When you look here https://github.com/dynatrace-oss/dynatrace-gcp-function/tree/all-services-preview#supported-google-cloud-services it says only a few services are supported.

I would suggest to add something that says, that you have out-of-the-box support for the list of services and a section saying how to add metrics from other services.

pgutkowski (approved):

PR #52, opened 2021-02-23
Feature/gcp function allows enabling http proxy separately for dt and gcp

Dawid-Kaszubski:
Function creates two separate client sessions. Base on `USE_PROXY` adds proxy support to client sessions.


pawelsiwek (approved):

mswiatkowska:
We should also handle proxy in run_docker.py - try_configure_dynatrace() and initial_check()

scripts/uninstall.sh
```diff
@@ -36,6 +36,7 @@ readonly ACTIVATION_SERVICES=$(yq r activation-config.yaml 'activation.metrics.s
 readonly PRINT_METRIC_INGEST_INPUT=$(yq r activation-config.yaml 'debug.printMetricIngestInput')
 readonly DEFAULT_GCP_FUNCTION_SIZE=$(yq r activation-config.yaml 'googleCloud.common.cloudFunctionSize')
 readonly SELF_MONITORING_DASHBOARD_NAME="dynatrace-gcp-function Self monitoring"
+readonly USE_PROXY=$(yq r $FUNCTION_ACTIVATION_CONFIG 'googleCloud.common.useProxy')
```
sq2gxo, line 39: Minor note: this setting is not used for uninstall, so I think it is not needed there

scripts/update.sh
```diff
@@ -39,6 +39,7 @@ readonly FUNCTION_GCP_SERVICES=$(yq r $FUNCTION_ACTIVATION_CONFIG 'activation.me
 readonly DASHBOARDS_TO_ACTIVATE=$(yq r -j -P $FUNCTION_ACTIVATION_CONFIG 'activation.metrics.services' | jq -r .[]) 
 readonly PRINT_METRIC_INGEST_INPUT=$(yq r $FUNCTION_ACTIVATION_CONFIG 'debug.printMetricIngestInput')
 readonly DEFAULT_GCP_FUNCTION_SIZE=$(yq r $FUNCTION_ACTIVATION_CONFIG 'googleCloud.common.cloudFunctionSize')
+readonly USE_PROXY=$(yq r $FUNCTION_ACTIVATION_CONFIG 'googleCloud.common.useProxy')
```
sq2gxo, line 42: Minor note: this setting is not used for update, so I think it is not needed there

sq2gxo (approved):

mswiatkowska (approved):

PR #67, opened 2021-04-01
Do not report zero counter values to Dynatrace

sq2gxo:


pgutkowski (approved):

benj5656:
We find zero values are useful for continuous monitoring, rather than just dropping the metric & reporting no data. 
No data also skews the way Davis raises alerts?

sq2gxo:
> We find zero values are useful for continuous monitoring, rather than just dropping the metric & reporting no data.
> No data also skews the way Davis raises alerts?

That is very good comment @benj5656 . This PR was initially a bugfix, as we already have logic to not send zero values to Dynatrace (src/lib/metric_ingest.py:287)
To reduce load - we don't want to send counter type zero delta values, but we will double check how Dynatrace charts and alerts are reacting to this change and then decide how to proceed.

sq2gxo:
We decided to not merge this PR

PR #72, opened 2021-04-07
APM-293893: Implement FastCheck for Logs

pgutkowski:
Introduced `LogsFastCheck` which prints out version and configuration flags for logs forwarder. Additionally it sends startup message to configured log ingest endpoint - which is verification of configuration and acts as a diagnostic information.
Additional changes:
- fix for outdated timestamps in integration test - test was passing but all messages were stashed due to too old timestamp
- Preserve asyncio loop in daemon thread to keep aiohttp working for logs forwarding

### DIFF
--- a/src/lib/fast_check.py
+++ b/src/lib/fast_check.py
@@ -1,6 +1,9 @@
 import asyncio
 import os
 import re
+import time
+from datetime import datetime
+from queue import Queue
 from typing import NamedTuple, List, Optional
 
 from urllib.parse import urljoin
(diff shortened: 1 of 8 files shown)


mswiatkowska (approved):

src/lib/fast_check.py
```diff
@@ -69,7 +83,7 @@ async def get_dynatrace_token_metadata(dt_session: ClientSession, context: Loggi
         return {}
 
 
-class FastCheck:
+class MetricsFastCheck:
```
mswiatkowska, line 86: Referring to our discussion about metrics fast check and disabled monitoring due to missing permissions (when an annotation is not created for the service account) - I checked and after your changes pod will be restarted until annotation is created, so after that monitoring will start working properly.

But I think we could change one log : from "Access to following projects: " to e.g. "There is no access to any projects. Check service account configuration." when 'projects' is empty. Actually I added this log, so I can change it in my PR.
What do you think? Would it be more readable for customer?

Dawid-Kaszubski (approved):

src/lib/fast_check.py
pgutkowski: Thank you for checking this scenario. Updating log message is neglible change, so I think that we can include it in your PR.

src/lib/instance_metadata.py
```diff
@@ -73,13 +73,14 @@ async def _init_(self) -> Optional[instance_metadata]:
 
             if not all(result is None for result in results):
                 audience = jwt.decode(results[4], verify=False) if results[4] else ''
-                return instance_metadata(
+                metadata = InstanceMetadata(
```
sq2gxo, line 76: metadata is only logged - previously was returned - is it intentional?

pgutkowski: It wasn't intentional, thank you for catching this.

mswiatkowska (approved):

sq2gxo (approved):

Dawid-Kaszubski (approved):

PR #76, opened 2021-04-15
Bugfix/gcp function dataflow dimension breaching the metric ingestion api dimension

Dawid-Kaszubski:
Matching the creation of dimensions to MINT API constraints. MINT API rejects dimension values longer than 250 chars, and dimension names longer than 100 chars.
Added unit tests and travis stage.


src/lib/metric_ingest.py
```diff
@@ -206,20 +208,34 @@ def extract_typed_value_key(time_serie):
     return typed_value_key
 
 
-def create_dimensions(time_serie: Dict) -> List[DimensionValue]:
+def create_dimension(name: str, value: str, context: LoggingContext = LoggingContext(None)) -> DimensionValue:
+    if len(name) > MAX_DIMENSION_NAME_LENGTH:
+        context.log("MINT rejects dimension names longer that {max_dim_name_length} chars. Dimension name \"{name}\" "
```
pgutkowski, line 213: [f-formatted strings](https://docs.python.org/3/tutorial/inputoutput.html) are generally used in this project, because these are more readable than `format` call

src/lib/metric_ingest.py
```diff
@@ -16,13 +16,15 @@
...
-from lib.context import Context, DynatraceConnectivity
+from lib.context import LoggingContext, Context, DynatraceConnectivity
 from lib.entities.ids import _create_mmh3_hash
 from lib.entities.model import Entity
 from lib.metrics import DISTRIBUTION_VALUE_KEY, Metric, TYPED_VALUE_KEY_MAPPING, GCPService, \
     DimensionValue, IngestLine
 
 UNIT_10TO2PERCENT = "10^2.%"
+MAX_DIMENSION_NAME_LENGTH = 100
+MAX_DIMENSION_VALUE_LENGTH = 250
```
pgutkowski, line 27: I think we should make these overridable with environment variables and add it to table in `HACKING.MD` - I don't really think any customer will need to use it, but that table acts as a documentation of known limits of MINT ingest too.

pgutkowski (changes requested):

pgutkowski (approved):

sq2gxo (approved):

PR #77, opened 2021-04-19
[APM-293900] support separate config entry for logs ingest Dynatrace endpoint

sq2gxo:
Added possibility to use sepatate endpoint for metrics (MINT API on cluster server) and logs (Generic Logs Ingest API on ActiveGate node)

### DIFF
--- a/activation-config.yaml
+++ b/activation-config.yaml
@@ -5,10 +5,8 @@ debug:
 googleCloud:
   common:
     # Name of Google Secret Manager Secret containing Dynatrace Access Key
-    dynatraceAccessKeySecretName: DYNATRACE_ACCESS_KEY 
-    # Name of Google Secret Manager Secret containing Dynatrace URL
-    dynatraceUrlSecretName: DYNATRACE_URL
-    # Name of the Service Account that should be created for Dynatrace Function to authenticate with Google Monitoring API 
+    dynatraceAccessKeySecretName: DYNATRACE_ACCESS_KEY
+    # Name of the Service Account that should be created for Dynatrace Function to authenticate with Google Monitoring API
     serviceAccount: dynatrace-gcp-service
     # Path to Google credentials file
     # googleApplicationCredentialsPath: ''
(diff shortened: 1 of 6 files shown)


src/lib/credentials.py
```diff
@@ -43,8 +45,13 @@ def get_dynatrace_api_key_from_env():
     return os.environ.get(_DYNATRACE_ACCESS_KEY_SECRET_NAME, None)
 
 
-def get_dynatrace_url_from_env():
-    return os.environ.get(_DYNATRACE_URL_SECRET_NAME, None)
+def get_dynatrace_log_ingest_url_from_env():
+    # TODO alternatively fetch from GCP secret
```
mswiatkowska, line 49: todo ?

activation-config.yaml
```diff
@@ -35,6 +35,8 @@ googleCloud:
     # Cron schedule for Cloud Scheduler instance
     schedulerSchedule: "* * * * *" 
   logs:
+    # Name of Google Secret Manager Secret containing Dynatrace logs ingest URL
```
mswiatkowska, line 38: I'm a little confused now - `activation-config.yaml` is used by the script for function deployment (not kubernetes setup) - as we added support for logs only for kubernetes setup why we have logs section in this file? What am I missing?

sq2gxo: You are right, I did too much changes :)

src/lib/credentials.py
sq2gxo: I didn't wanted to introduce too much changes in this PR. I'll create separate task to implement it as I suspect it will require more code changes

mswiatkowska: If I understand correctly fetch_secret() checks if env variable is set (if yes returns the value if not secret is fetched from secret manager) -> get_dynatrace_log_ingest_url_from_env() could just call fetch_secret() ?

activation-config.yaml
sq2gxo: It seem entire `logs:` section is not needed - I'm removing it

src/lib/credentials.py
sq2gxo: I confirmed that fetch_secret() is (officially) used only by function. I'll remove TODO, it is not needed

mswiatkowska (approved):

pgutkowski (approved):

PR #84, opened 2021-04-29
APM-292502: GCP logs self monitoring

mswiatkowska:


src/lib/logs/dynatrace_client.py
```diff
@@ -43,6 +47,7 @@ def send_logs(context: LogsContext, logs: List[Dict]):
             encoded_body_bytes = batch.encode("UTF-8")
             context.log('Log ingest payload size: {} kB'.format(round((len(encoded_body_bytes) / 1024), 3)))
 
+            self_monitoring.all_requests += 1
```
pgutkowski, line 50: IMO we are missing reporting of size of sent ingest (the number we are printing in line 48) and optionally the number of sent log entries, although this one can be inferred from number of ackes pubsub messages.

src/lib/logs/dynatrace_client.py
```diff
@@ -54,18 +59,40 @@ def send_logs(context: LogsContext, logs: List[Dict]):
...
+                    self_monitoring.dynatrace_connectivity.append(DynatraceConnectivity.InvalidInput)
+                elif status == 401:
+                    self_monitoring.dynatrace_connectivity.append(DynatraceConnectivity.ExpiredToken)
+                elif status == 403:
+                    self_monitoring.dynatrace_connectivity.append(DynatraceConnectivity.WrongToken)
+                elif status == 404 or status == 405:
+                    self_monitoring.dynatrace_connectivity.append(DynatraceConnectivity.WrongURL)
+                elif status == 413 or status == 429:
+                    self_monitoring.dynatrace_connectivity.append(DynatraceConnectivity.TooManyRequests)
+                    raise HTTPError(log_ingest_url, 429, "Dynatrace throttling response")
```
pgutkowski, line 72: Did you test this `elif` branch? My IDE marks this HTTPError call as missing 2 additional constructor arguments, and indeed the constructor accepts 5 arguments. The same applies to line `75`, and it

src/lib/logs/log_sfm_metric_descriptor.py
```diff
@@ -0,0 +1,148 @@
...
+LOGS_SUBSCRIPTION_ID_LABEL_DESCRIPTOR = {
+    "key": "logs_subscription_id",
+    "valueType": "STRING",
+    "description": "Subscription id of log sink pubsub subscription"
+}
+
+LOG_SELF_MONITORING_ALL_REQUESTS_METRIC_DESCRIPTOR = {
+    "type": LOG_SELF_MONITORING_ALL_REQUESTS_METRIC_TYPE,
+    "valueType": "INT64",
+    "metricKind": "GAUGE",
```
pgutkowski, line 39: I think we may have a problem here with `gauge` metrics and the way `context.function_name` is extracted:
````python
        self.function_name = os.environ.get("FUNCTION_NAME", "Local")
````
Metrics self monitoring code was written for function deployment and never updated for Kuberenetes deployment. It was not an issue, as metrics have to run in single container, but for logs we are going to use horizontal scaling heavily. I suspect that containers don't have `FUNCTION_NAME` env variable, which would make instances overwrite gauge values of other instances. We should refine (or at least check) code in `src/lib/context.py:83` to make it work correctly.

File: src/lib/logs/log_self_monitoring.py [hunk 4]
```diff
@@ -0,0 +1,277 @@
...
+        aggregated_sfm.records_with_too_long_content += sfm.records_with_too_long_content
+        aggregated_sfm.dynatrace_connectivity.extend(sfm.dynatrace_connectivity)
+        aggregated_sfm.processing_time += sfm.processing_time
+        aggregated_sfm.sending_time += sfm.sending_time
+    return aggregated_sfm
+
+
+def put_sfm_into_queue(sfm_queue: Queue, sfm: LogSelfMonitoring, logging_context: LoggingContext):
+    try:
+        sfm_queue.put(sfm, True, SENDING_WORKER_EXECUTION_PERIOD_SECONDS + 1)
```
pgutkowski, line 69: I don't think we should block processing/sending workers on `sfm_queue`, we can grind processing to the halt for several seconds this way. Logging exception should be enough

pgutkowski (changes requested):
A few comments, but you did great job in this PR @mswiatkowska!

HACKING.md
```diff
@@ -33,6 +33,7 @@ Worker function execution can be tweaked with environment variables. In Google F
 | USE_PROXY | Depending on value of this flag, function will use proxy settings for either Dynatrace, GCP API or both. Allowed values: `ALL`, `DT_ONLY`, `GCP_ONLY` |  |
 | IMPORT_DASHBOARDS | Import predefined dashboards for selected services. Allowed values: `yes`, `no` | `yes` |
 | IMPORT_ALERTS | Import predefined alerting rules (inactive by default) for selected services. Allowed values: `yes`, `no`| `yes` |
+| SELF_MONITORING_ENABLED | Send custom metrics to GCP to diagnose quickly if your gcp-log-forwarder processes and sends logs to Dynatrace properly | False |
```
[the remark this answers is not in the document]
mswiatkowska: Thanks for catching that, copy/paste mistake...

src/lib/logs/dynatrace_client.py
mswiatkowska: Added both, however calculating number of sent logs entries need to be improved after improvement for ack messages (to not acknowledge messages when request to Dynatrace fails)

src/lib/logs/logs_processor.py
```diff
@@ -82,19 +91,22 @@ def _create_dt_log_payload(context: LoggingContext, message_data: str) -> Option
...
+        self_monitoring.too_old_records += 1
         return None
 
     content = parsed_record.get(ATTRIBUTE_CONTENT, None)
     if content:
         if not isinstance(content, str):
             parsed_record[ATTRIBUTE_CONTENT] = json.dumps(parsed_record[ATTRIBUTE_CONTENT])
-        if len(parsed_record[ATTRIBUTE_CONTENT]) >= _CONTENT_LENGTH_LIMIT:
-            parsed_record[ATTRIBUTE_CONTENT] = parsed_record[ATTRIBUTE_CONTENT][:_CONTENT_LENGTH_LIMIT]
+        CONTENT_LENGTH_LIMIT = get_int_environment_value("DYNATRACE_LOG_INGEST_CONTENT_MAX_LENGTH", 8192)
```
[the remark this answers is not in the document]
mswiatkowska: I had a problem with imports and env variables in integration test, forgot to handle it. Finally I moved CONTENT_LENGTH_LIMIT to log_forwarder_variables module.

src/lib/logs/dynatrace_client.py
mswiatkowska: You're right 2 arguments are missing - added them (although I was sure I tested it and seemed to be working;) ).

src/lib/logs/log_sfm_metric_descriptor.py
mswiatkowska: It seems like GCP doesn't allow to overwrite time series - one can be sent each 10 seconds and they need to be written in order. I added new metric label - container_name - it will prevent time series from rejecting. Unfortunately it's not possible to get cluster_name that's why I left function_name as is.

tests/integration/test_integration.py
```diff
@@ -80,6 +82,10 @@ def cleanup():
 def setup_env(monkeypatch):
     for variable_name in system_variables:
         monkeypatch.setenv(variable_name, system_variables[variable_name])
+    reload(log_forwarder_variables)
```
pgutkowski, line 85: Nice one, I didn't know this trick

pgutkowski (approved):

src/lib/logs/log_self_monitoring.py
```diff
@@ -0,0 +1,309 @@
...
+        aggregated_sfm.processing_time += sfm.processing_time
+        aggregated_sfm.sending_time += sfm.sending_time
+        aggregated_sfm.log_ingest_payload_size += sfm.log_ingest_payload_size
+        aggregated_sfm.sent_logs_entries += sfm.sent_logs_entries
+    return aggregated_sfm
+
+
+def put_sfm_into_queue(context: LogsContext):
+    try:
+        context.sfm_queue.put(context.self_monitoring, False, SENDING_WORKER_EXECUTION_PERIOD_SECONDS + 1)
```
pgutkowski, line 56: (Minor) this unsused timeout value `SENDING_WORKER_EXECUTION_PERIOD_SECONDS + 1` is confusing at glance, you can use `put_nowait(item)` instead of `put` to be more explicit that this call is not blocking

src/dashboards/dynatrace-gcp-function-log-self-monitoring
```diff
@@ -0,0 +1,266 @@
+{
```
sq2gxo, line 1: filename is missing .json extension

sq2gxo: and as this is GCP (not Dynatrace) dashboard, maybe we should put it into separate directory, e.g. `src/dashboards/gcp'? What do you think?

sq2gxo (approved):

src/dashboards/dynatrace-gcp-function-log-self-monitoring
mswiatkowska: There is already dashboard for metric self monitoring, that's why I put it there. For me it's not necessary to create separate directory, maybe someone else could share an opinion.

sq2gxo: right, there is already another non-Dynatrace dashboard here, I'm fine with keeping it there

PR #85, opened 2021-04-30
GCP Function HTTPS_PROXY and HTTP_PROXY environment variables are def…

Dawid-Kaszubski:
…ined in config map and helm template

### DIFF
--- a/k8s/dynatrace-gcp-function.yaml
+++ b/k8s/dynatrace-gcp-function.yaml
@@ -99,15 +99,15 @@ spec:
               name: dynatrace-gcp-function-config
               key: IMPORT_ALERTS
         - name: HTTP_PROXY
-            valueFrom:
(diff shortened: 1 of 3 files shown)


sq2gxo (approved):

k8s/dynatrace-gcp-function.yaml
```diff
@@ -95,6 +98,16 @@ spec:
             configMapKeyRef:
               name: dynatrace-gcp-function-config
               key: IMPORT_ALERTS
+        - name: HTTP_PROXY
+            valueFrom:
```
mswiatkowska, line 102: There is incorrect indentation before 'valueFrom' - it should be the same as in the line above.

mswiatkowska (approved):

pgutkowski (approved):

PR #92, opened 2021-05-10
change helm artifact file name

sq2gxo:


mswiatkowska (approved):

sq2gxo:
As pointed out by @pgutkowski it is better to stay with name without version - a we will also manage official documentation here: https://www.dynatrace.com/support/help/technology-support/cloud-platforms/google-cloud-platform/monitor-gcp-services-with-dynatrace/deploy-in-k8-container/

PR #106, opened 2021-05-26
APM-297036: Batch self monitoring data

pgutkowski:


pgutkowski:
This PR introduces removing docker-build stage as well. Decision has been consulted with @pawelsiwek. There is no point in building container on every build, since we have test suite now and Dockerhub throttling is making our CI fail

dtwwolinski (approved):

mswiatkowska (approved):

PR #107, opened 2021-05-27
Create helm deployment script

mswiatkowska:
- the deployment script uses only the most important properties, there are no properties like: printMetricIngestInput, serviceUsageBooking, useProxy - AFAIK they are used rarely, only in specific cases and still they can be changed in config map. @pgutkowski , @pawelsiwek  what do you think? 
- needed to escape comma in gcp-services argument used in helm install command - comma is a special character in --set property
- I didn't add additional check for kubernetes connection - `kubectl get namespace dynatrace` returns exception with message if there is no connection: `"Unable to connect to the server: x509: certificate signed by unknown authority"` - imo it's enough.


pgutkowski:
Ad 1. We can skip `printMetricIngestInput` as it was intended to be debug tool, but we should support `serviceUsageBooking` and `useProxy` as these are valid use cases, and best practice is not to manually update resources created by Helm because then upgrading with new version of chart can overwrite the changes.
Ad 2. Please research if it would be possible to annotate k8s service account in chart template, instead of making it separate step.

pgutkowski (approved):

pgutkowski (dismissed):

dtwwolinski (approved):
Thanks for adding double [quotes](https://github.com/koalaman/shellcheck/wiki/SC2086) and removing point 8. 👍

pgutkowski (approved):

pgutkowski (approved):

PR #112, opened 2021-06-08
APM-292701: Mapping Audit Logs

pgutkowski:
Minor change to metadata engine itself - discussion with @pawelsiwek lead to conclusion we should apply `audit_logs` rules to logs matched with other rules, and apply them last, to overwrite any attributes introduced by other rules.

### DIFF
--- a/src/lib/logs/jmespath.py
+++ b/src/lib/logs/jmespath.py
@@ -66,7 +66,7 @@ class MappingCustomFunctions(functions.Functions):
         if not proto_code:
             return "Succeeded"
         else:
-            return "Failure." + self.proto_error_code_to_string_dict.get(proto_code, "")
+            return "Failed." + self.proto_error_code_to_string_dict.get(proto_code, "")
 
 
 JMESPATH_OPTIONS = jmespath.Options(custom_functions=MappingCustomFunctions())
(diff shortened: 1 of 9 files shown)


src/lib/logs/jmespath.py
```diff
@@ -41,5 +41,32 @@ def _func_if(self, condition, if_true_expression, if_false_expression, node_scop
...
+        15: "DataLoss",
+        16: "Unauthenticated"
+    }
+
+    @functions.signature({'types': []})
+    def _func_status_from_proto_code(self, proto_code):
+        if not proto_code:
+            return "Succeeded"
+        else:
+            return "Failure." + self.proto_error_code_to_string_dict.get(proto_code, "")
```
mswiatkowska, line 69: In Azure failures start with "failed." (logs I've got during test)  e.g. failed.notfound, failed.badrequest - maybe we should use the same prefix? 
Do we know what prefix AWS uses?

pgutkowski: Yeah, I intended to use the same prefix as in Azure, but I got it wrong

mswiatkowska (approved):

Dawid-Kaszubski (approved):

PR #114, opened 2021-06-11
APM-306738: helm chart improvements

mswiatkowska:
Added checksum/config to Deployment - allows to roll the pods of the deployment every time the configMap is changed


src/lib/logs/log_self_monitoring.py
```diff
@@ -25,7 +25,7 @@
 from lib.context import LoggingContext, LogsSfmContext, DynatraceConnectivity, LogsContext
 from lib.credentials import create_token, get_dynatrace_log_ingest_url_from_env
 from lib.instance_metadata import InstanceMetadata
-from lib.logs.log_forwarder_variables import LOGS_SUBSCRIPTION_PROJECT, LOGS_SUBSCRIPTION_ID, \
+from lib.logs.log_forwarder_variables import GCP_PROJECT, LOGS_SUBSCRIPTION_ID, \
```
pgutkowski, line 28: In case if anybody already used our Helm template and pulls new docker image, we should fallback to `LOGS_SUBSCRIPTION_PROJECT` if `GCP_PROJECT` is not set

src/lib/fast_check.py
```diff
@@ -39,7 +39,7 @@
     "DYNATRACE_LOG_INGEST_REQUEST_MAX_SIZE",
     "DYNATRACE_TIMEOUT_SECONDS",
     "DYNATRACE_LOG_INGEST_EVENT_MAX_AGE_SECONDS",
-    "LOGS_SUBSCRIPTION_PROJECT",
```
pgutkowski, line 42: I'd leave the old flag in case if new version image is used in old deployment.

pgutkowski (changes requested):

build/helm-package.sh
```diff
@@ -30,5 +30,7 @@ if [ ! -f "$FILE" ]; then
 fi
 
 sudo mv linux-amd64/helm /usr/local/bin/helm
-RESULT_STR=$(helm package ./k8s/helm-chart/dynatrace-gcp-function -d ./artefacts)
-mv ./${RESULT_STR##* } ./artefacts/dynatrace-gcp-function.tgz
\ No newline at end of file
+mkdir ./helm-deployment-package
```
dtwwolinski, line 16: Since we are not using the **helm** anymore there's no need to create wget it and check if it's available.

dtwwolinski (changes requested):

dtwwolinski (approved):

Dawid-Kaszubski (approved):

sq2gxo (approved):

pgutkowski (approved):

PR #117, opened 2021-06-24
E2E test - part 1: deployment 

sq2gxo:


tests/e2e/deployment-test.sh
```diff
@@ -0,0 +1,145 @@
...
+sed -i '/^dynatraceLogIngestUrl:/c\dynatraceLogIngestUrl: "'${DYNATRACE_LOG_INGEST_URL}'"' ${VALUES_FILE}
+sed -i '/^dynatraceUrl:/c\dynatraceUrl: "'${DYNATRACE_URL}'"' ${VALUES_FILE}
+sed -i '/^logsSubscriptionId:/c\logsSubscriptionId: "'${PUBSUB_SUBSCRIPTION}'"' ${VALUES_FILE}
+sed -i '/^requireValidCertificate:/c\requireValidCertificate: "false"' ${VALUES_FILE}
+
+gcloud container clusters get-credentials "${K8S_CLUSTER}" --region us-central1 --project ${GCP_PROJECT_ID}
+
+cd ./e2e_test || exit 1
+chmod +x ./deploy-helm.sh
+./deploy-helm.sh --service-account e2e-test-gcp-function-sa --role-name e2e_test_gcp_function --quiet || exit 1
```
pgutkowski, line 108: I'm wondering if we shouldn't create unique `service-account` and `role` for each pipeline run (e.g. by adding build id as suffix to names). With fixed names, pipeline will create them only on the first run, and confirm they exist on proceeding runs. This way we are not testing the most crucial path - creating these resources.

sq2gxo: I agree, but there this need to be carefully checked whether there is a ris on hitting GCP limits.
I'm mostly concerned on custom roles hard limit which is 300. Need to verify if custom roles in deleted state are counted against limit.

pgutkowski: According to this doc: https://cloud.google.com/iam/docs/creating-custom-roles#deleting-custom-role deleted role is not counted toward the limit

sq2gxo: after 7 days ;)
"After 7 days, the role is scheduled for permanent deletion. At this point, the role no longer counts towards the limit of 300 custom roles per organization or 300 custom roles per project."
Still we should be safe as build will run only for PRs and in master

scripts/deploy-helm.sh
```diff
@@ -96,6 +98,11 @@ while (( "$#" )); do
                 shift; shift
             ;;
 
+            "-q" | "--quiet")
```
mswiatkowska, line 101: (minor) can we add -q argument to print_help() function?

mswiatkowska (approved):

mswiatkowska (approved):

pgutkowski (approved):

PR #118, opened 2021-06-30
Containerized ag for gcp

dtwwolinski:
- Added new option in values to install Active Gate in k8s as separate Pod.
- Pod consists of Secret, StatefullSet and ClusterIP
- Added _helpers file with functions to extract specific values:
-- environmentID of given cluster.
-- URL to Cluster registry to pull.
-- Creates a secret for docker.
-- Creates a URL for log ingest URL.
- Checking in case the autoDeploy is ON if PaaS Token is filled.


k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
```diff
@@ -0,0 +1,108 @@
...
+  name: dynatrace-ag-gcpmon
+  labels:
+    app: dynatrace-ag-gcpmon
+  namespace: dynatrace
+spec:
+  selector:
+    matchLabels:
+      app: dynatrace-ag-gcpmon
+  replicas: 1
+  serviceName: dynatrace-ag-gcpmon
```
pgutkowski, line 51: Did you come up with any solution for long startup time and synchronization with Dyantrace - GCP Function pods?

k8s/helm-chart/dynatrace-gcp-function/Chart.yaml
```diff
@@ -16,4 +16,13 @@ apiVersion: v2
...
+home: https://github.com/dynatrace-oss/dynatrace-gcp-function
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 1.4.0
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+# It is recommended to use it with quotes.
+appVersion: "1.17.0"
```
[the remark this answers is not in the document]
dtwwolinski: You are right. Generally, the chartVersion should be updated each time you change the chart. Not sure how Travis should handle this. On the other hand, the app version should be filled by Travis. Maybe the solution would be to just have the same version for both handled by Travis. I'll create an improvement for that since it does exceed the scope of this story. The file will be reverted.

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
```diff
@@ -0,0 +1,108 @@
...
+  {{- if (.Values.activeGate.autoDeploy) }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: dynatrace-activegate-gcpmon-router
+  namespace: dynatrace
+spec:
+  type: ClusterIP
+  selector:
+    app: dynatrace-ag-gcpmon
```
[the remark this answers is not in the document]
dtwwolinski: Good idea. I'll change the name of AG pod.  As it comes to the second one I'm not sure since by listing all resources u can strictly say what it is. For example **dynatrace-gcp-function-sa** or **dynatrace-gcp-function-secret**.

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
dtwwolinski: Working on it. For now, I'll try an ``` initConteiner ``` and check for RUNNING state of  AG pod.

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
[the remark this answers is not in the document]
pgutkowski: okay, maybe we could replace the `router` with `service` to explicitly say what kind of Kubernetes resource it is?

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
```diff
@@ -0,0 +1,108 @@
...
+        checksum/secret: {{ include (print $.Template.BasePath "/secret.yaml") . | sha256sum }}
+    spec:
+      containers:
+        - name: dynatrace-ag-gcpmon
+          image: {{ template "activeGateImage" . }}
+          imagePullPolicy: Always
+          ports:
+            - name: ag-https
+              containerPort: 9999
+          env:
```
mswiatkowska, line 69: I'm wondering... if we had two AGs installed in the same gcp project (image downloaded from the same tenant), they would have the same id?

k8s/helm-chart/dynatrace-gcp-function/values.yaml
```diff
@@ -27,11 +27,22 @@ deploymentType: "all"
...
 
+#dynatraceLogIngestUrl ActiveGate endpoint used to ingest logs to Dynatrace, for example: https://environemnt-active-gate-url:9999/e/environment-id"
+#used ONLY when the activeGate.autoDeploy is false
+dynatraceLogIngestUrl: ""
+
+activeGate:
+  #autoDeploy if set to true the Active Gate will be deployed in k8s. In that case also the activeGate.dynatracePaasToken must be filled.
+  autoDeploy: true
+  #autoDeployPort port used by Active Gate pod on which AG is exposed.
+  autoDeployPort : 9999
```
pgutkowski, line 42: Do we have a specific use-case for k8s service custom port? This `values.yaml` file keeps growing and feedback from customers have been that they feel overwhelmed by amount of options. I'd remove `autoDeployPort` if it's not absolutely necessary.

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
sq2gxo: that is corner case but still possible, @mswiatkowska do you remember how it is handled for Azure logs forwared? if I remember correctly Azure Resource Group name is used as one of SEED components?

k8s/helm-chart/dynatrace-gcp-function/values.yaml
```diff
@@ -27,11 +27,22 @@ deploymentType: "all"
...
 # subscription id of log sink pubsub subscription
 logsSubscriptionId: ""
 
+#dynatraceLogIngestUrl ActiveGate endpoint used to ingest logs to Dynatrace, for example: https://environemnt-active-gate-url:9999/e/environment-id"
+#used ONLY when the activeGate.autoDeploy is false
+dynatraceLogIngestUrl: ""
+
+activeGate:
+  #autoDeploy if set to true the Active Gate will be deployed in k8s. In that case also the activeGate.dynatracePaasToken must be filled.
+  autoDeploy: true
```
sq2gxo, line 40: name and logic of new patrameter is not consistant wit Azure logs forwarder.
In Azure AG it is reverse :
`            "--use-existing-active-gate")
                USE_EXISTING_ACTIVE_GATE=true`
https://github.com/dynatrace-oss/dynatrace-azure-log-forwarder/blob/master/deployment/dynatrace-azure-logs.sh#L107

k8s/helm-chart/dynatrace-gcp-function/values.yaml
dtwwolinski: Removed it.

k8s/helm-chart/dynatrace-gcp-function/values.yaml
dtwwolinski: Changed to match other clouds.

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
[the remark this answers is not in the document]
dtwwolinski: Changed

k8s/helm-chart/dynatrace-gcp-function/values.yaml
```diff
@@ -27,11 +27,22 @@ deploymentType: "all"
...
+  #autoDeployPort port used by Active Gate pod on which AG is exposed.
+  autoDeployPort : 9999
+  #dynatracePaasToken PaaS token, generated in Dynatrace cluster.
+  dynatracePaasToken: ""
+
 # METRICS VALUES - REQUIRED for 'metrics' and 'all' deployment
 # Dynatrace environment endpoint, for example: https://environment-id.live.dynatrace.com
 dynatraceUrl: ""
```
mswiatkowska, line 48: dynatraceUrl is not used only in metrics deployment anymore, please move it to common values with changed description when it should be set

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
dtwwolinski: I've added a .Values.clusterName to generate a new ID

k8s/helm-chart/dynatrace-gcp-function/templates/_helpers.tpl
```diff
@@ -0,0 +1,67 @@
...
+
+{{/*
+activeGateUrl in case of the .Values.activeGate.useExisting is
+     false -> the  .Values.dynatraceLogIngestUrl is NOT used as an Active Gate URL. For log ingest. The URL is pointing to autodeployed AG inside k8s cluster.
+     true -> the  .Values.dynatraceLogIngestUrl IS used as an Active Gate URL. For log ingest.
+*/}}
+{{- define "activeGateUrl" }}
+{{- if eq .Values.activeGate.useExisting "false"}}
+  {{- $envid := (include "environmentID" .)}}
+  {{- printf "https://dynatrace-activegate-gcpmon-clusterip.dynatrace.svc.cluster.local:9999/e/%s"  $envid }}
```
karolkieglerski, line 58: please use `Release.Namespace` instead of hardcoded namespace in URL

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
dtwwolinski: The init Container worked as expected.

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
```diff
@@ -0,0 +1,108 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+{{- if or (eq .Values.deploymentType "logs") (eq .Values.deploymentType "all") }}
+  {{- if eq .Values.activeGate.useExisting "false" }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: dynatrace-activegate-gcpmon-clusterip
+  namespace: dynatrace
```
karolkieglerski, line 20: also putting this is in each template is not good practice... because during installation you also setting namespace so some parts can be install in different namespace

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
```diff
@@ -0,0 +1,108 @@
...
+            - name: DT_ID_SEED_ENV_ID
+              value: {{ template "environmentID" . }}
+            - name: DT_ID_SEED_CLUSTER_NAME
+              value: {{ .Values.clusterName }}
+          resources:
+            requests:
+              memory: 1024Mi
+              cpu: 1
+            limits:
+              memory: 1024Mi
```
karolkieglerski, line 85: are you sure that request should be the limit? if it is a java then the limit should be at least 20% above

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
```diff
@@ -0,0 +1,108 @@
...
+              cpu: 1
+          livenessProbe:
+            httpGet:
```
karolkieglerski, line 88: if any customer will have an istio we will have a bug here

k8s/helm-chart/dynatrace-gcp-function/templates/template.yml
```diff
@@ -228,4 +228,13 @@ spec:
             path: /health
             port: 8081
           initialDelaySeconds: 10
+      {{- if (eq .Values.activeGate.useExisting "false") }}
+      initContainers:
+        - name: init-active-gate-service
+          image: busybox:1.33.1
+          command: [ 'sh', '-c', "until nslookup dynatrace-activegate-gcpmon-clusterip.dynatrace.svc.cluster.local; do echo waiting for AG service; sleep 2; done" ]
```
karolkieglerski, line 235: if you are on the same namespace then use short link: 
```
until nslookup dynatrace-activegate-gcpmon-clusterip
```
instead of:
```
until nslookup dynatrace-activegate-gcpmon-clusterip.dynatrace.svc.cluster.local
```

k8s/helm-chart/dynatrace-gcp-function/templates/template.yml
```diff
@@ -228,4 +228,13 @@ spec:
...
+        - name: init-active-gate
+          image: radial/busyboxplus:curl
+          command: [ 'sh', '-c', 'until [ $(curl  https://dynatrace-activegate-gcpmon-clusterip.dynatrace.svc.cluster.local:9999/rest/state -k)  == "RUNNING" ];do echo "Checking for Running Status";echo "Response = $AG_RESPONE.";sleep 2;done;echo "AG running"' ]
```
karolkieglerski, line 238: same as above

k8s/helm-chart/dynatrace-gcp-function/values.yaml
```diff
@@ -84,6 +94,8 @@ importDashboards: "true"
 # Import predefined alerting rules for selected services. Allowed values: "true"/"yes", "false"/"no"
 importAlerts: "true"
 # configuration file for GCP services - uncommented services will be monitored
+#clusterName Cluster Name, should be empty as it's filed by setup.sh.
```
karolkieglerski, line 97: what does it mean? I didn't find setting this up in setup.sh

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
pawelsiwek: why so? if request = limit we have Guaranteed  QoS and this is safe for AG>..

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
pawelsiwek: Namespace for Dynatrace resources as of now must not be different than dynatrace. (all resources including Operator, Agent are there). Having this in mind, dynamic namespace will rather obfuscate the deployment code :)

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
karolkieglerski: if we have request and limit on 1k then java should have around 800Mi because java + OS will use only 1k so if java will take around 1k then k8s will kill pod but java will have no chances to clear garbage collector ex. and work again without restart

k8s/helm-chart/dynatrace-gcp-function/values.yaml
dtwwolinski: Fixed file reference.

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
karolkieglerski: and also in different way if we are using only 500Mi MAX all the time (not in spikes) so we not need that memory... if our pod want more but not above the limit then k8s will evict other pods to allow our pod to work

dtwwolinski: In Ag repo, in yamls for AG deployment, the request limits are set to 250M. IMHO 1024Mi is more than enough.

k8s/helm-chart/dynatrace-gcp-function/templates/template.yml
dtwwolinski: Good to know :)

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
karolkieglerski: so limit as 1k is ok or to enough but request should be set to 250 or 500

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
karolkieglerski: ok so we should remove it from deployment command because we have hardcoded namespace in each template

pawelsiwek: which deployment command do U refer to?

karolkieglerski: https://github.com/dynatrace-oss/dynatrace-gcp-function/pull/118/files#diff-8c092cd24fc76c45c928dfac92e5f6593cf7854a3eb844c1b8dee57312eecfecR316

k8s/helm-chart/dynatrace-gcp-function/templates/_helpers.tpl
dtwwolinski: Nice to know, since the URL can be shorter I'll use the ```https://dynatrace-activegate-gcpmon-clusterip```

k8s/helm-chart/dynatrace-gcp-function/values.yaml
dtwwolinski: moved

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
dtwwolinski: Agreed. Changed to 512MB

k8s/helm-chart/dynatrace-gcp-function/templates/_helpers.tpl
karolkieglerski: if it is only call from that one namespace you can use:
```
https://dynatrace-activegate-gcpmon-clusterip:9999/e/%s
```

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
pawelsiwek: memory mgmt. is optimized on AG level. AG itself sets heap usage to 50% available limit memory (default java limit is 25%). Having this in mind still we can benefit from Guaranteed QoS.

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-pod.yaml
pawelsiwek: ok forgot about helm :) actually I've double checked what we do in other helm charts currently and yes namespace is dynamic there https://github.com/Dynatrace/helm-charts/blob/master/dynatrace-operator/templates/Kubernetes/operator/deployment-operator.yaml . So actually I agree @karolkieglerski  with Your comment to make it variable here as well.

karolkieglerski (dismissed):

pgutkowski (approved):

karolkieglerski (approved):

mswiatkowska (approved):

pgutkowski (approved):

PR #125, opened 2021-07-07
Validate ActiveGate endpoint health before proceeding with installation

sq2gxo:


scripts/deploy-helm.sh
```diff
@@ -129,7 +129,7 @@ while (( "$#" )); do
     esac
 done
 
-readonly GCP_PROJECT=$(helm show values ./dynatrace-gcp-function --jsonpath "{.gcpProjectId}")
+GCP_PROJECT=$(helm show values ./dynatrace-gcp-function --jsonpath "{.gcpProjectId}")
```
sq2gxo, line 132: removed readonly as this var could be modified in line 140

sq2gxo:
I decided to not fail installation on no connectivity to AG - the issue is that script can be run from anywhere and it may not have connectivity to AG host - as it turned out in E2E test

scripts/deploy-helm.sh
```diff
@@ -201,13 +201,25 @@ if [[ $DEPLOYMENT_TYPE == all ]] || [[ $DEPLOYMENT_TYPE == logs ]]; then
...
+    exit 1
+  fi
 fi
 
 if [[ $CREATE_AUTOPILOT_CLUSTER == "Y" ]]; then
   SELECTED_REGION=$(gcloud config get-value compute/region 2>/dev/null)
   if [ -z "$SELECTED_REGION" ]; then
     echo
-    echo "\e[93mWARNING: \e[37mDefault region not set. Set default region by running 'gcloud config set compute/region <REGION>'."
+    echo - e "\e[93mWARNING: \e[37mDefault region not set. Set default region by running 'gcloud config set compute/region <REGION>'."
```
karolkieglerski, line 222: @sq2gxo use `echo -e` instead `echo - e`

mswiatkowska (approved):

pgutkowski (approved):

karolkieglerski (approved):

Dawid-Kaszubski (approved):

PR #129, opened 2021-07-08
Apm 311119 improve deployment script output on successfull deployment

mswiatkowska:


scripts/deploy-helm.sh
```diff
@@ -286,3 +286,14 @@ fi
...
 fi
+
+if [[ $DEPLOYMENT_TYPE == all ]]; then
+  LOG_MONITORING_URL="${DYNATRACE_URL}/ui/log-monitoring?query=cloud.provider%3D%22gcp%22"
+  LOG_VIEWER="Log Viewer: $LOG_MONITORING_URL"
+fi
+
+if [[ $DEPLOYMENT_TYPE == logs ]] || [[ $DEPLOYMENT_TYPE == all ]]; then
+  echo
+  echo "- Check logs in Dynatrace in 5 min. ${LOG_VIEWER}" >${CMD_OUT_PIPE}
```
karolkieglerski, line 297: if you will have only `logs` then variable `LOG_VIEWER` will be not exists here

karolkieglerski (changes requested):

scripts/deploy-helm.sh
karolkieglerski: explained

karolkieglerski (dismissed):

dtwwolinski (approved):

scripts/deploy-helm.sh
mswiatkowska: Yes, we can show the link only when DYNATRACE_URL is set (for 'all' deployment type), for 'logs' deployment type it's empty deliberately

dt-be (approved):

scripts/deploy-helm.sh
```diff
@@ -286,3 +286,14 @@ fi
 if [[ $DEPLOYMENT_TYPE == metrics ]] || [[ $DEPLOYMENT_TYPE == all ]]; then
   echo "kubectl -n dynatrace logs -l app=dynatrace-gcp-function -c dynatrace-gcp-function-metrics"  > ${CMD_OUT_PIPE}
 fi
+
+if [[ $DEPLOYMENT_TYPE == all ]]; then
```
sq2gxo, line 290: please add comment explaining why we cannot construct LOG_VIEWER url for deployment type = 'logs'

scripts/deploy-helm.sh
```diff
@@ -389,11 +389,24 @@ echo "- 7. Install dynatrace-gcp-function with helm chart in $CLUSTER_NAME"
...
   echo "kubectl -n dynatrace logs -l app=dynatrace-gcp-function -c dynatrace-gcp-function-logs"  > ${CMD_OUT_PIPE}
 fi
 
 if [[ $DEPLOYMENT_TYPE == metrics ]] || [[ $DEPLOYMENT_TYPE == all ]]; then
   echo "kubectl -n dynatrace logs -l app=dynatrace-gcp-function -c dynatrace-gcp-function-metrics"  > ${CMD_OUT_PIPE}
+fi
+
+if [[ $DEPLOYMENT_TYPE == all ]]; then
+  # To build a Log viewer link we need a Dynatrace url which is set only for 'metrics' and 'all' deployment types.
+  # For 'logs' deployment type (where ActiveGate url is used) we are not able to build the link - LOG_VIEWER is empty then.
```
dtwwolinski, line 403: By default, we are using the containerized AG, thus dynatraceURL is known since we are using it to pull AG image from the given cluster. So to summarize the Log_viewer URL **can't** be constructed only for the deployment type 'logs'  with the use of existing AG enabled.

karolkieglerski (dismissed):

dtwwolinski (approved):

scripts/deploy-helm.sh
```diff
@@ -189,7 +189,7 @@ readonly DYNATRACE_URL_REGEX="^https:\/\/[-a-zA-Z0-9@:%._+~=]{1,256}[\/]{0,1}$"
 readonly ACTIVE_GATE_TARGET_URL_REGEX="^https:\/\/[-a-zA-Z0-9@:%._+~=]{1,256}\/e\/[-a-z0-9]{1,36}[\/]{0,1}$"
 
 GCP_PROJECT=$(helm show values ./dynatrace-gcp-function --jsonpath "{.gcpProjectId}")
-readonly DEPLOYMENT_TYPE=$(helm show values ./dynatrace-gcp-function --jsonpath "{.deploymentType}")
+DEPLOYMENT_TYPE=$(helm show values ./dynatrace-gcp-function --jsonpath "{.deploymentType}")
```
karolkieglerski, line 192: why you removed `readonly` ?

mswiatkowska: removed because in line 217 we can override it (however it's a case when deployment type is empty in values.yaml)

Dawid-Kaszubski (approved):

karolkieglerski (approved):

PR #130, opened 2021-07-09
Sed replacement with yq

karolkieglerski:
note: yq not found on travis


mswiatkowska (dismissed):

tests/e2e/deployment-test.sh
```diff
@@ -47,7 +47,8 @@ while (( "$#" )); do
     esac
 done
 
-
+# Install YQ
+curl -sSLo yq "https://github.com/mikefarah/yq/releases/download/3.4.0/yq_linux_amd64" && chmod +x yq && sudo mv yq /usr/local/bin/yq
```
pgutkowski, line 51: We are using old version of yq in `setup.sh`, but is there anything forcing us to use this specific version here?

sq2gxo (dismissed):

tests/e2e/deployment-test.sh
pgutkowski: Docs apply to GCP Cloud Function deployment, which we hope to deprecate altogether soon. In E2E test we are using deployment with Helm and we can use different tools.

pgutkowski (approved):

sq2gxo (approved):

mswiatkowska (approved):

PR #132, opened 2021-07-12
Apm 311715 handle any message format

pgutkowski:


mswiatkowska (approved):

src/lib/logs/log_forwarder.py
```diff
@@ -132,6 +132,9 @@ def perform_flush(worker_state: WorkerState,
                 context.self_monitoring.sent_logs_entries += len(worker_state.jobs)
                 context.self_monitoring.log_ingest_payload_size += display_payload_size
                 send_batched_acks(subscriber_client, subscription_path, worker_state.ack_ids)
+        elif worker_state.ack_ids:
+            # Send ACKs if processing all messages has failed
```
sq2gxo, line 136: Will it ACK if messages were correctly processed but Dynatrace AG endpoint is not available?

sq2gxo: looking at the code - I think it will not, wchich is correct :)

src/lib/logs/log_self_monitoring.py
```diff
@@ -228,6 +230,21 @@ def create_self_monitoring_time_series(sfm: LogSelfMonitoring, context: LogsSfmC
                     "value": {"int64Value": sfm.records_with_too_long_content}
                 }]))
 
+    if sfm.publish_time_fallback_records:
+        time_series.append(
+            create_time_serie(
```
sq2gxo, line 235: we should add new timeseries to dynatrace-gcp-function-log-self-monitoring.json dashboard

sq2gxo (commented):
SFM dashboard should be updated with new SFM metric

src/lib/logs/log_forwarder.py
pgutkowski: You're right- if any message will be correctly processed, then interpreter will follow execution branch on line 122, and ACK only if logs have been correctly sent to Dynatrace.

src/lib/logs/log_self_monitoring.py
pgutkowski: Good catch, I've missed it

mswiatkowska (approved):

PR #139, opened 2021-07-19
Apm 304721 e2e test metrics

mswiatkowska:


.travis.yml
```diff
@@ -82,6 +84,7 @@ jobs:
       - GCP_PROJECT_ID="dynatrace-gcp-extension"
       - GCR_NAME="gcr.io/${GCP_PROJECT_ID}/dynatrace-gcp-function"
       - E2E=true
+      - CLOUD_FUNCTION_NAME="sample_app"
```
dtwwolinski, line 87: Please change the function name to something meaningful :) for example e2e_test_sample_app

tests/e2e/deployment-test.sh
```diff
@@ -96,7 +96,7 @@ writerIdentity=$(gcloud logging sinks describe "${LOG_ROUTER}" --format json | j
 gcloud pubsub topics add-iam-policy-binding "${PUBSUB_TOPIC}" --member ${writerIdentity} --role roles/pubsub.publisher > /dev/null 2>&1
 
 # Create E2E Sample App
-gcloud functions deploy sample_app \
+gcloud functions deploy "${CLOUD_FUNCTION_NAME}" \
 --runtime python37 \
```
dtwwolinski, line 100: This is out of the scope of this task, but please consider changing the test runtime to python38 as this is the version that we are using.

mswiatkowska: It's just a sample app which produces logs for e2e tests, it's not related with our processing logs function

karolkieglerski: you must change also here https://github.com/dynatrace-oss/dynatrace-gcp-function/pull/139/files#diff-ae94410f78e789a88f96202c7f4a8ef10ef33ec028de15ecf6ecc80b4c9bcfadR162 because it will fail

tests/e2e/deployment-test.sh
```diff
@@ -159,7 +159,7 @@ kubectl -n dynatrace get pods
 
 # Generate load on GC Function
 for i in {1..5}; do
-  curl "https://us-central1-${GCP_PROJECT_ID}.cloudfunctions.net/sample_app?deployment_type=${DEPLOYMENT_TYPE}&build_id=${TRAVIS_BUILD_ID}" \
+  curl "https://us-central1-${GCP_PROJECT_ID}.cloudfunctions.net/e2e_test_sample_app?deployment_type=${DEPLOYMENT_TYPE}&build_id=${TRAVIS_BUILD_ID}" \
```
karolkieglerski, line 162: @mswiatkowska if you are using variable `"${CLOUD_FUNCTION_NAME}"` use it also here instead of hardcoded name

scripts/deploy-helm.sh
```diff
@@ -32,7 +32,7 @@ check_if_parameter_is_empty()
 
 check_api_token() {
   URL="$1"
-  if RESPONSE=$(curl -k -s -X POST -d "{\"token\":\"$DYNATRACE_ACCESS_KEY\"}" "$URL/api/v2/apiTokens/lookup" -w "<<HTTP_CODE>>%{http_code}" -H "accept: application/json; charset=utf-8" -H "Content-Type: application/json; charset=utf-8" -H "Authorization: Api-Token $DYNATRACE_ACCESS_KEY"); then
+  if RESPONSE=$(curl -k -s -X POST -d "{\"token\":\"$DYNATRACE_ACCESS_KEY\"}" "$URL/api/v2/apiTokens/lookup" -w "<<HTTP_CODE>>%{http_code}" -H "accept: application/json; charset=utf-8" -H "Content-Type: application/json; charset=utf-8" -H "Authorization: Api-Token $DYNATRACE_ACCESS_KEY" --connect-timeout 20); then
```
karolkieglerski, line 35: after reducing this timeout can by case why tests are not passing right now

mswiatkowska: yeah it's possible, I will add sleep inside the test for logs, timeout should stay for those requests

dtwwolinski (approved):

mswiatkowska:
ready for the review

tests/e2e/logs/test_e2e.py
```diff
@@ -13,6 +13,8 @@
 #     limitations under the License.
 
 import os
+import time
```
karolkieglerski, line 16: this `import` looks like not used

karolkieglerski (approved):

tests/e2e/logs/test_e2e.py
mswiatkowska: removed

karolkieglerski (approved):

sq2gxo (approved):

PR #145, opened 2021-07-28
[APM-293895] Proxy support for gcp log forwarder

dtwwolinski:
- updated the config map to have configuration for proxies,
- for now this setup works only if an existing AG is used.


k8s/helm-chart/dynatrace-gcp-function/templates/template.yml
```diff
@@ -77,11 +77,13 @@ spec:
             configMapKeyRef:
               name: dynatrace-gcp-function-config
               key: SERVICE_USAGE_BOOKING
+
         - name: USE_PROXY
           valueFrom:
             configMapKeyRef:
               name: dynatrace-gcp-function-config
               key: USE_PROXY
+        {{- if or (eq .Values.useProxy "ALL")  (eq .Values.useProxy "DT_ONLY") (eq .Values.useProxy "GCP_ONLY")}}
```
sq2gxo, line 86: expression can be simplified to `if .Values.useProxy` - empty value of useProxy means - no proxy

dtwwolinski: Simplified

karolkieglerski:
working locally and from Docker not tested from k8s cluster yet

Dawid-Kaszubski (dismissed):

scripts/deploy-helm.sh
```diff
@@ -263,7 +250,11 @@ else
   exit 1
 fi
 
-  check_proxy_settings
+if [ -n "$USE_PROXY" ]; then
+  if [ -z "$HTTP_PROXY" ] || [ -z "$HTTPS_PROXY" ]; then
+    echo -e "\e[91mERROR: \e[37mThe useProxy is set, please fill httpProxy or httpsProxy in your values file"
```
sq2gxo, line 253: should we continue deployment with this error? or should we exit in case of incorrect config?

HACKING.md
```diff
@@ -54,6 +54,9 @@ Worker function execution can be tweaked with environment variables. In Google F
 | DYNATRACE_LOG_INGEST_REQUEST_MAX_SIZE | Max size in bytes of single payload to logs ingest endpoint. If it surpasses server limit, payload will be rejected with 413 code | 1048576 (1 mb) |
 | DYNATRACE_LOG_INGEST_EVENT_MAX_AGE_SECONDS | Determines max age of forwarded log event. Should be the same or lower than on cluster | 1 day |
 | GCP_PROJECT | GCP project of log sink pubsub subscription | |
+| USE_PROXY | Depending on value of this flag, function will use proxy settings for either Dynatrace, GCP API or both.
+| HTTP_PROXY | Set the proxy address. To be used in conjunction with USE_PROXY |  |
```
sq2gxo, line 58: is it possible to have both specified? if yes, conditions in helm template may not work corretly

karolkieglerski: yes it is possible and is also recommended. Why? I fix it to cover that case

scripts/deploy-helm.sh
karolkieglerski: hmmmm, as you think. From my perspective if you not set url it will work without proxy

sq2gxo (dismissed):

HACKING.md
karolkieglerski: fixed

k8s/helm-chart/dynatrace-gcp-function/templates/function-deployment.yml
```diff
@@ -210,6 +214,52 @@ spec:
             configMapKeyRef:
               name: dynatrace-gcp-function-config
               key: SELF_MONITORING_ENABLED
+        {{- if and (.Values.useProxy) (ne .Values.useProxy "default")}}
```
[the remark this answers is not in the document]
karolkieglerski: fixed

karolkieglerski (approved):

k8s/helm-chart/dynatrace-gcp-function/templates/_helpers.tpl
```diff
@@ -59,9 +59,29 @@ activeGateUrl in case of the .Values.activeGate.useExisting is
...
+{{- end }}
 
+{{/*
+noProxyUrls is used in case of the .Value.useProxy is set ALL, DT_ONLY or GCP_ONLY
+  In a case the AG will be installed the communication between pods doesn't need a proxy.
+  In a case we are using an existing AG and the proxy should be used only for GCP then the communication between Function and AG must not use the proxy.
+  In a case we are using an existing AG and the proxy should be used only for DT.
+*/}}
+{{- define "noProxyUrls"}}
+{{- if eq .Values.activeGate.useExisting "false" -}}
```
mswiatkowska, line 76: With this condition we don't handle the case when a client choose new AG deployment and useProxy "DT_ONLY".
I agree that when new AG is created during installation, communication doesn't need a proxy but still a client can put "DT_ONLY" in values file -> then our proxy will work unpredictably/incorrect -> we won't set `googleapis.com` in noProxyUrls so we will use proxy for gcp requests.
IMO we need to adjust this condition to handle such case somehow. It's important to notice that adding `googleapis.com` to noProxyUrls (when DT_ONLY is set) is equivalent to not using proxy at all...

@karolkieglerski , @sq2gxo , @pawelsiwek what do you think? Any ideas?:)

k8s/helm-chart/dynatrace-gcp-function/templates/active-gate-statefulset.yaml
```diff
@@ -77,6 +78,13 @@ spec:
               value: {{ template "environmentID" . }}
             - name: DT_ID_SEED_CLUSTER_NAME
               value: {{ .Values.clusterName }}
+            {{- if  or (eq .Values.useProxy "DT_ONLY") (eq .Values.useProxy "ALL") }}
+            - name: DT_INTERNAL_PROXY
+              valueFrom:
+                configMapKeyRef:
+                  name: dynatrace-gcp-function-config
+                  key: HTTP_PROXY
```
Dawid-Kaszubski, line 86: HTTPS also needed

karolkieglerski: Proxy used for communication with Dynatrace server

Value: percent encoded proxy URI

Example:

--internalProxy=https://username:password@hostname:port

Dawid-Kaszubski (approved):

k8s/helm-chart/dynatrace-gcp-function/templates/_helpers.tpl
karolkieglerski: @dtwwolinski

sq2gxo (approved):

mswiatkowska (approved):

karolkieglerski (approved):

PR #146, opened 2021-07-28
Apm 315215 cloud sql logs attributes extraction

sq2gxo:


tests/unit/extraction_rules/test_audit_activity.py
```diff
@@ -332,6 +333,7 @@
         ATTRIBUTE_GCP_REGION: 'europe-north1',
         ATTRIBUTE_GCP_PROJECT_ID: 'dynatrace-gcp-extension',
         ATTRIBUTE_GCP_RESOURCE_TYPE: 'cloudsql_database',
+        ATTRIBUTE_GCP_INSTANCE_ID: 'dynatrace-gcp-extension:pawel-001-mysql',
```
karolkieglerski, line 336: pointing to specific private pawel instance is not good idea I think. Better will be to create separate instance only for tests.

src/config_logs/cloud_sql.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "cloud_sql",
+  "displayName": "Cloud SQL",
```
karolkieglerski, line 3: this name doesn't mean which kind of DB it is, mysql, postgresql or sql server

tests/unit/extraction_rules/test_cloud_sql.py
```diff
@@ -0,0 +1,55 @@
...
+  "logName": "projects/dynatrace-gcp-extension/logs/cloudsql.googleapis.com%2Fmysql-slow.log",
+  "receiveTimestamp": "2021-07-26T12:08:26.686970384Z"
+}
+expected_output = {
+    ATTRIBUTE_CLOUD_PROVIDER: 'gcp',
+    ATTRIBUTE_CLOUD_REGION: 'europe-north1',
+    ATTRIBUTE_GCP_REGION: 'europe-north1',
+    ATTRIBUTE_GCP_PROJECT_ID: 'dynatrace-gcp-extension',
+    ATTRIBUTE_GCP_RESOURCE_TYPE: 'cloudsql_database',
+    ATTRIBUTE_GCP_INSTANCE_ID: 'dynatrace-gcp-extension:test-001-mysql',
```
karolkieglerski, line 46: the same as above

tests/unit/extraction_rules/test_cloud_sql.py
```diff
@@ -0,0 +1,55 @@
...
+    ATTRIBUTE_TIMESTAMP: timestamp,
+    ATTRIBUTE_CONTENT: 'ON DUPLICATE KEY UPDATE master_time=UTC_TIMESTAMP(6);',
+    ATTRIBUTE_DT_LOGPATH: 'projects/dynatrace-gcp-extension/logs/cloudsql.googleapis.com%2Fmysql-slow.log'
```
karolkieglerski, line 49: only here we have information that this is for mysql

tests/unit/extraction_rules/test_audit_activity.py
sq2gxo: it is just unit test string, not using any real instance

src/config_logs/cloud_sql.json
sq2gxo: yes, beacause we don't have anythink in the GCP logs for cloud_sql that we can use to detect engine; this config will be common for all eninges of cloud sql

tests/unit/extraction_rules/test_audit_activity.py
dtwwolinski: Although I understand your reasoning @sq2gxo, I agree with @karolkieglerski on that, since Pawel or someone might remove this instance  because he wouldn't know that it's used for testing

sq2gxo: as we discussed today - it is just a string value but I've changed it anyway

tests/unit/extraction_rules/test_cloud_sql.py
sq2gxo: we try to can extract this info and put into db.system property, but I'll follow up on this in next PR

karolkieglerski (approved):

Dawid-Kaszubski (approved):

tests/unit/extraction_rules/test_cloud_sql.py
mswiatkowska: Do another logs ('general' / 'error' ) also have the same format name - 'mysql-<log_type>.log'?
We could put this log type to log.source property too.

mswiatkowska (approved):

PR #148, opened 2021-08-03
APM-313824 add exception throttling

equardo:
This is incomplete - the throttling logic is there and I added caller names to metadata-engine log calls and any error() or exception()  calls. The problem is that LoggingContest's log function is called over 100 times in the project seemingly without obvious distinction between kinds of calls. Me looking through all of them to decide which to throttle and which not would be counterproductive as well as possibly full of mistakes. My question is if someone could maybe take over this PR/branch to finish it or is it maybe not worth the time at all?


Dawid-Kaszubski:
1. Changed approach from caller to message key, cause when using *args argument, caller argument was added to a message.
2. Moved throttling message dict to `LoggingContext` class. Message occurrences are now counted per `LoggingContext` instance.  `LoggingContext` is created for every function run, so occurrences are in a single function run context.
3. Changed logging to throttled for exceptions and errors, which may have a lot of occurrences.

karolkieglerski (approved):

src/lib/context.py
```diff
@@ -65,6 +68,28 @@ def exception(self, *args):
...
         traceback.print_exc()
 
+    def t_error(self, *args):
+        """
+        Prints error with throttling limit per message. Limit per message set in LOG_THROTTLING_LIMIT_PER_CALLER
+        Use for potentially frequent log.
+        :param args:
+        :return:
+        """
+        if args and self.__check_if_message_exceeded_limit(args[-1]):
```
equardo, line 78: Just checking, but isn't this approach not very optimal? If an error were to be repeatedly called, but each time with slightly different arguments, it would keep counting it separately.

equardo: As to the issue of passing the caller id to be printed, you could just pass all but the last argument from t_error to error

Dawid-Kaszubski: You're right, that with this approach messages with different arguments will be logged. I've decided to count occurrences per message on purpose. 
We get rid of log spam of already logged errors and exceptions and we keep logging messages with different arguments. These messages might be crucial for troubleshooting and I believe that troubleshooting is the main purpose of logging.

karolkieglerski:
after your changes script `generate_metadata.py` starts failing on:
```
Failed to map units: {'By.s', 'GiBy.s', 'GBy.s'}
```

karolkieglerski (changes requested):
fix generate_metadata.py script and also check generate_ddu_estimation.py

karolkieglerski (approved):

src/lib/context.py
```diff
@@ -90,6 +116,21 @@ def log(self, *args):
 
         print(f"{timestamp_utc_iso} {context_section} : {message}")
 
+    def __check_if_message_exceeded_limit(self, message: str):
+        log_calls_performed = self.throttled_log_call_count.get(message, 0)
```
sq2gxo, line 120: I'm afraid of memory usage in case of very high number of similar messages but with small difference (e.g. containing timestamp).
log_calls_performed map should have expiration policy - be somehow  periodically cleared of old entries

sq2gxo (changes requested):

src/lib/context.py
```diff
@@ -65,6 +68,28 @@ def exception(self, *args):
         self.error(*args)
         traceback.print_exc()
 
+    def t_error(self, *args):
+        """
+        Prints error with throttling limit per message. Limit per message set in LOG_THROTTLING_LIMIT_PER_CALLER
```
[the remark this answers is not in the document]
Dawid-Kaszubski: LoggingContexts are created per single polling run. So it's impossible to miss these kinds of errors.

src/lib/context.py
Dawid-Kaszubski: LoggingContexts are created per single polling run and the map is a field of this object. So map will be recreated for every single polling.
For LoggingContext created at the start of the container - it's used only on start, so there is no need to throttle logs for it.

Dawid-Kaszubski: I was wondering about some type of list of timestamps for recurring logs, but I feel like it's overkill. Especially, that we have never had problems with lots of logs for GCP.

sq2gxo (approved):

mswiatkowska (approved):

PR #149, opened 2021-08-06
Apm 309526 align gcp metric dimension names with gcp log semantic attributes

dt-be:
Ready for review.

### CompatibilityMode=False / None - for any new customer, default
reporting only new dims, e.g. gcp.project.id (project_id not there)
image
image
image

### CompatibilityMode=True - for customers who had previous setups
reporting both old and new dims, e.g. both gcp.project.id and project_id reported
image
image

New dashboard - only new dashboard working with new data (when adding a service after an update)
To test the dashboard, I've added BeluTestin12 to the dashboards' names to force it to be recognized as a new one and pushed to Dt cluster. 
Note, that as of Dt cluster version 223 & 224 dashboards are correctly uploaded, but not correctly published (set to preset=true), since a bug in RestAPI (APM-322917, https://dynatrace.slack.com/archives/GMLF457LL/p1630587604053200). 

image

Old dashboards - for already added services, old dashboards should still be working (old dashboards use old dims, which are still reported), giving consistent view of the data. Hard for me to demonstrate this, as jxw has data from multiple sources.
Note: old dashboards SHOULD be in env since when the service was added before: BUT this might not be true if there was anything wrong in the API for uploading/publishing dashboards when service was added (the bug already mentioned in 223+ or the one in June)


src/config/gce_zone_network_health.yml
```diff
@@ -15,10 +15,19 @@ gcp:
...
   - value: resource.labels.zone
     id: zone
+    compatibilityModeOnly: true
+  - value: resource.labels.zone
+    id: gcp.region
   - value: resource.labels.region
     id: region
+    compatibilityModeOnly: true
+  - value: resource.labels.region
+    id: gcp.region
```
dt-be, line 30: TODO BŁ duplicate, keep only one as gcp.region - zone, because it's more detailed?

dt-be:
References for checking what should and what shouldn't go to gcp.instance.name:
https://cloud.google.com/monitoring/api/resources
https://cloud.google.com/asset-inventory/docs/resource-name-format

dt-be:
FAILED tests/unit/test_activation_config.py::test_filtering_missing_configs
`> load_supported_services(context, [])`

I think this started failing because of 2bc7ceef (APM-309526 fix load_supported_services(selected_featuresets = []) usages: ddu & metadata generation - load all if no whitelist_exists)
@pgutkowski I think this is about what I told you that I fixed it to get my `test_compatibility_mode.py` running, as it was also used and failing those:
```
generate_ddu_estimation.py
generate_metadata.py
```
You've mentioned the behaviour could have changed because of some filtering implemented. I'm not here next two weeks, could you or @mswiatkowska help whoever takes this over to navigate what it should be like in the end? Should it allow loading absolutely all service configs, or is it not allowed anymore for some reasons and could only be run in test, and non-prod?

Temporarily commenting out f659d31d

src/main.py
```diff
@@ -276,13 +276,14 @@ def load_supported_services(context: LoggingContext, selected_featuresets: List[
                     service_name=service_yaml.get("service", "None")
                     featureSet=service_yaml.get("featureSet", "None")
                     # If whitelist of services exists and current service is not present in it, skip
+                    whitelist_exists = selected_featuresets is not None and selected_featuresets.__len__() > 0
                     if activation_config:
                         activation_config_feature_sets = activation_config.get(service_name, {}).get("featureSets", [])
-                        should_skip = featureSet not in activation_config_feature_sets
+                        should_skip = whitelist_exists and featureSet not in activation_config_feature_sets
```
mswiatkowska, line 282: selected_featuresets should be used only as a fallback (in else section, probably I should have updated the comment) - it's a previous solution (in the config we could specify list of gcp services to monitor).
Currently with new deployments we check if feature set from service config is present in activation_config, if not then skip it.

activation-config.yaml
```diff
@@ -49,6 +49,11 @@ googleCloud:
     scheduler: dynatrace-gcp-schedule
     # Cron schedule for Cloud Scheduler instance
     schedulerSchedule: "* * * * *"
+    # Compatibility mode - concerns the change in reported dimension names. Allowed values: "true"/"yes", "false"/"no".
+    # Should be false for all new setups. Meant to allow a gradual transition for setups that already have historical metrics data.
+    # If true, report deprecated metric dimensions (flagged as compatiblityModeOnly=true) along with their new equivalents:
+    # project_id > gcp.project.id | region, location, zone, metrics.label.region > gcp.region
+    metrics: false
```
mswiatkowska, line 56: 'compatibilityMode' instead of 'metrics'?

src/config/apigee.googleapis.com_ProxyV2.yml
```diff
@@ -21,8 +21,14 @@ gcp:
...
+  - value: resource.labels.location
+    id: gcp.region
   - value: resource.labels.proxy_name
     id: proxy_name
+    compatibilityModeOnly: true
+  - value: resource.labels.proxy_name
+    id: gcp.instance.name
   - value: resource.labels.runtime_version
     id: runtime_version
   - value: resource.labels.instance_id
```
mswiatkowska, line 34: Shouldn't we create new dimension `gcp.instance.id` here too? or maybe we have already created such dimension in the code (couldn't find). @pgutkowski could you maybe confirm?

dt-be:
Moving into the comment:

## 1. configs + load

### DONE:
* configs have new dimensions with compatibilityModeOnly=True
* on configs being loaded (load_supp_services > ... > new Dimension) dimensions get filtered according to compatibilityMode env_var
* test for loading of configs (unit/test_compatibility_mode) + checks en masse if nothing skipped through

### TODO:
* metrics.ODIN_COMP...MODE TODO BŁ - should probably be init once, but test need to patch it differently then
*  ~~manual review in PR: of changes accross all config commits:
**  make sure only one of location/region + zone made it to new field gcp.region
**  make sure only one _name field made it to new field gcp.instance.name~~ BŁ DONE

* [nice to have] cleanup fields in Dimension value Class - try to make fewer of them, I tried to bring it all in one place to see how it is used in different places, next step would be to try and reduce the number of fields to some common subset

## 2. changes to how are dt dimension keys produced

### DONE
* checked by regex in configs, that magic on dimension.value in create_dimension matches dimension.id
** confirmed by PawelG that this was created automatically, so +1 to those 2 should match
* tried to verify no breaking changes in dims produced - the dims produced before/after changes to create_dimension - list all dims produced per single polling before/after: [ReportedDims Before & After.xlsx](https://github.com/dynatrace-oss/dynatrace-gcp-function/files/6948220/ReportedDims.Before.After.xlsx)

* run locally (local_test), works fine

### TODO:
* ~~!!! review: more verification to double check no breaking changes to dims!!!
** verify output visually who can?
** verify on cluster/jxw no new dims?
** verify on cluster/jxw history continuity?
** anything that occurs to you~~ DONE BŁ: run locally before&after results: [ReportedDims Before & After.pdf](https://github.com/dynatrace-oss/dynatrace-gcp-function/files/7056807/ReportedDims.Before.After.pdf)

* integration test > test_integration_metric.py
* verify e2e test pass
* make build - fix test_activation_config.py
~~* make build - fix test_integration_metric.py -~~ fixed, but still to consider if something to add here. It catched something, so that's nice, [nice to have] have another scenario with compatibilityMode=true


## 3. dashboards
### DONE
~~* moved exisiting dashboards to > compatibilityMode~~
~~* created new non-compatibilityMode dashboards > changed all dims there~~
~~* config changes to point to both compatibilityMode & nonCompMode dashboard~~

### TODO
~~* verify script uploading dashboards for function: run bash setup.sh~~
~~* verify script uploading dashboards for kubernetes: run run_docker.py locally~~
~~* verify what we wanted - had some mindfuck concerning how, so might be the scripts do something useless~~ 
I don't know what lack of memory possessed me here, I've now fixed the dashboards to be a single set of dashboards including both dimension sets
* test if it works with real data - for old dims reported, for new dims reported (compMode=false) and for both (compatibilityMode=true)
** script uploading dashboards for function: run bash setup.sh
** script uploading dashboards for kubernetes: run run_docker.py locally > lib.configure_dynatrace.ConfigureDynatrace.import_dashboards

dt-be:
Note nr 1, for customers who had previous setups imho we should either:

* make it very explicit in the Release notes that customers with existing setups should set CompatibilityMode=True if they want to see continuous data, or
* somehow make it set to true on update/helm-upgrade? I'm not sure if this is possible - I could try to investigate that, but I don't think it was originally discussed in scope of this task, so let me know what do you think
Otherwise, a customer who updates and misses setting the flag to true will only have data with old dims up to the update point, and after the update point only with new dims.

dt-be:
Note as to the dashboards/shareSettings restAPI bug:
Dashboards can only be published manually in the UI if you first get the uuids of the individual dashboards from container logs or via the dashboards RestAPI.
Once the dashboard shareSettings RestAPI is fixed, they still won't be published (the script won't retry that if they are already uploaded). I'd seriously consider changing the script to try to publish the unpublished ones. Otherwise, you probably need to send any customer explicitly **to the RestAPI to delete them** - as their owner is Dynatrace Open Source and they are not typically visible in the UI.

dt-be:
I've only tested full GCP setup in K8 version. Let me know if you see the need to test function setup as well.
E2E test for this would be nice, too, but imo that's definitely out of scope of this task.

scripts/setup.sh
```diff
@@ -105,6 +105,7 @@ readonly HTTP_PROXY=$(yq e '.googleCloud.common.httpProxy' $FUNCTION_ACTIVATION_
 readonly HTTPS_PROXY=$(yq e '.googleCloud.common.httpsProxy' $FUNCTION_ACTIVATION_CONFIG)
 readonly IMPORT_DASHBOARDS=$(yq e '.googleCloud.common.importDashboards' $FUNCTION_ACTIVATION_CONFIG)
 readonly IMPORT_ALERTS=$(yq e '.googleCloud.common.importAlerts' $FUNCTION_ACTIVATION_CONFIG)
+readonly COMPATIBILITY_MODE=$(yq r $FUNCTION_ACTIVATION_CONFIG 'googleCloud.common.compatibilityMode')
```
karolkieglerski, line 108: `r` is invalid for `yq 4+`
```
Error: unknown command "r" for "yq"
```

tests/unit/test_activation_config.py
```diff
@@ -36,7 +36,7 @@ def test_filtering_config_blank_when_activation_config_missing():
     assert any(elem.name == "pubsub_subscription" and elem.monitoring_filter == '' for elem in config)
     assert any(elem.name == "pubsub_snapshot" and elem.monitoring_filter == '' for elem in config)
 
-
-def test_filtering_missing_configs():
-    config = load_supported_services(context, [])
-    assert len(config) == 0
+# TODO BŁ NOMERGE! commenting out to have e2e test run
```
karolkieglerski, line 39: so if it is ready to review this should be removed

karolkieglerski:
after merging this we should rewrite whole extensions @sq2gxo

tests/unit/test_activation_config.py
dt-be: 👍 I need to clear something for this in review when Maria comes back tomorrow

src/lib/metrics.py
```diff
@@ -57,16 +68,36 @@ def to_string(self) -> str:
 class Dimension:
     """Represents singular dimension."""
 
+    id: Text
+    value: Text
     dimension: Text
     source: Text
+    key_for_get_func_create_entity_id: Text
```
wojtasskorcz, line 75: Is it used anywhere?

src/lib/metrics.py
```diff
@@ -57,16 +68,36 @@ def to_string(self) -> str:
...
+    key_for_create_entity_id: Text
+    key_for_fetch_metric: Text
+    key_for_send_to_dynatrace: Text
 
     def __init__(self, **kwargs):
```
wojtasskorcz, line 80: This class, and especially its constructor seems overly complicated compared to how much information it actually carries. I see you replaced all usages of `dimension` and `source` in code except for L43 in `ids.py`. Maybe you should also replace it there and get rid of the `dimension` and `source` fields altogether and just use your newly introduced `key_for_*` fields?

Aside from their lack of usage in the rest of the codebase, these `dimension` and `source` fields introduce an unnecessary layer of indirection in the constructor when it comes to calculating the values of `key_for_*` fields. I think this constructor could be made simpler if we removed them

karolkieglerski (changes requested):
after your changes script `generate_metadata.py` starts failing on:
```
Failed to map units: {'By.s', 'GiBy.s', 'GBy.s'}
```
fix generate_metadata.py script and also check generate_ddu_estimation.py

tests/unit/test_activation_config.py
dt-be: Adapted after discussion with @mswiatkowska

dt-be:
> after your changes script `generate_metadata.py` starts failing on:
> 
> ```
> Failed to map units: {'By.s', 'GiBy.s', 'GBy.s'}
> ```
> 
> fix generate_metadata.py script and also check generate_ddu_estimation.py

As discussed in slack, nothing changed here, same behavior in master for both generate_metadata/ddu. 

The only change from me is that they can now also be run for all services without setting locally env_var ACTIVATION_CONFIG with explicitly listing all services.

src/config/apigee.googleapis.com_ProxyV2.yml
dt-be: After discussion with both Maria & @pawelsiwek, I'm also adding those fields that were there in odin spec here:
https://bitbucket.lab.dynatrace.org/projects/ODIN/repos/dt-semantic-conventions/browse/doc/semconv/cloud_gcp.md
image
but not present in jira when it was created:
* `gcp.instance.id` in configs whenever applicable
* `type` equal to service name & hardcoded in code (not configs)

Actual changes still need some confirmation with Pawel on Monday 😢, as I have some discrepancies with my choice of fields for .id and .name and his in one wiki he has sent to me.

src/lib/metric_ingest.py
```diff
@@ -119,6 +119,25 @@ async def log_invalid_lines(context: MetricsContext, ingest_response_json: Dict,
...
+class DtDimensionsMap:
+    def __init__(self) -> None:
+            self.dt_dimensions_set_by_source_dimension = {}
+
+    def add_label_mapping(self, source_dimension, dt_target_dimension):
+        all_dt_dims_for_source_dim = self.dt_dimensions_set_by_source_dimension.get(source_dimension, set())
+        all_dt_dims_for_source_dim.add(dt_target_dimension)
+        self.dt_dimensions_set_by_source_dimension[source_dimension] = all_dt_dims_for_source_dim
+
+    def create_dimensions(self, target_dimension_list: List, dt_label_if_unmapped: str, source_label: str, dim_value, context):
```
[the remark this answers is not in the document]
dt-be: I didn't like the output param either, but was scared to touch it at the time, because I didn't have automatic test then, so this was a welcome reminder 👍 

I was targeting brevity on the calling side back then - it's now bit more verbose now, but I'm happy with it as it still seems clear enough after the change - take a look guys if you like it better or have more thoughts about it.

src/lib/metrics.py
dt-be: fixed

src/lib/metrics.py
dt-be: Cleaned this up a bit, tests seem fine so far. It should look better than what you've commented on here, not 100% sure if better than what was here originally - so opinions welcome

wojtasskorcz: Looks much better now! I'm not checking if all the values stayed the same after the revamp but I trust that if you have tests for it then we're good

wojtasskorcz (approved):

karolkieglerski (approved):

Dawid-Kaszubski (approved):

sq2gxo (approved):

PR #163, opened 2021-08-31
[APM-297894] Improve diagnostics of GCP metric ingest function

dtwwolinski:
- moved fast check dynatrace to print token metadata and URL during execution.


src/lib/fast_check.py
```diff
@@ -96,6 +96,25 @@ async def get_dynatrace_token_metadata(dt_session: ClientSession, context: Loggi
...
 
+async def check_dynatrace(logging_context: LoggingContext, project_id, dt_session: ClientSession, dynatrace_url, dynatrace_access_key):
+    try:
+
+        if not dynatrace_url or not dynatrace_access_key:
+            logging_context.log(f'ERROR No Dynatrace secrets: DYNATRACE_URL, DYNATRACE_ACCESS_KEY for project: {project_id}.'
+                                     f'Add required secrets to Secret Manager.')
+            return None
+        logging_context.log(f"Using [DYNATRACE_URL] Dynatrace endpoint: {dynatrace_url}")
+        logging_context.log(f'Using [DYNATRACE_ACCESS_KEY]: {dynatrace_access_key[0:dynatrace_access_key.rfind(".")]}.')
```
sq2gxo, line 107: Customers can still also use tokens in old format - this line will reveal the token.
Characters until first dot are usually the same in new format:  'dt0c01'

src/lib/fast_check.py
```diff
@@ -76,7 +76,7 @@ def valid_dynatrace_scopes(token_metadata: dict):
 async def get_dynatrace_token_metadata(dt_session: ClientSession, context: LoggingContext, dynatrace_url: str, dynatrace_api_key: str, timeout: Optional[int] = 2) -> dict:
     try:
         response = await dt_session.post(
-            url=f"{dynatrace_url.rstrip('/')}/api/v1/tokens/lookup",
+            url=f"{dynatrace_url.rstrip('/')}/api/v2/apiTokens/lookup",
```
Dawid-Kaszubski, line 79: Does `v2/apiTokens` support tokens in old format?

dtwwolinski: It does. I've chosen but the output of v2 is more readable.

dtwwolinski: Reverted to v1 as the name of the token is the only needed thing.

src/lib/fast_check.py
```diff
@@ -96,6 +96,36 @@ async def get_dynatrace_token_metadata(dt_session: ClientSession, context: Loggi
         return {}
 
 
+def format_dynatrace_access_key(dynatrace_access_key :str):
+    if len(dynatrace_access_key) >= 7:
+        if dynatrace_access_key.startswith("dt0c01.") and len(dynatrace_access_key) >= 14:
```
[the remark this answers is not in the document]
sq2gxo: we cannot assume the lenght, it is good to be defensive against IndexError

src/lib/fast_check.py
```diff
@@ -168,8 +178,6 @@ async def execute(self) -> FastCheckResult:
             if all(result is not None for result in results):
                 ready_to_monitor.append(project_id)
 
-        await self._check_dynatrace(get_project_id_from_environment())
```
sq2gxo, line 171: I think that check was executed for K8S runtime and should stay here also

src/lib/fast_check.py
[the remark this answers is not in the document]
karolkieglerski: @sq2gxo didn't catch your answer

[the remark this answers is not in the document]
sq2gxo: ignore my answer, I didn't catch you question :)

sq2gxo (approved):

Dawid-Kaszubski (approved):

PR #165, opened 2021-08-31
APM-319866 fix string metrics metadata generations

karolkieglerski:
@dtmawo for what it was needed? right now it fails scripts because it skipping all `STRING` metrics


karolkieglerski:
@pgutkowski give me information that we not support STRING metrics

PR #166, opened 2021-09-01
Fix issue with creating redundant dashboards

sq2gxo:


src/lib/self_monitoring.py
```diff
@@ -300,18 +300,18 @@ def create_self_monitoring_time_series(context: MetricsContext) -> Dict:
...
-        )
-        if response.status <= 202:
-            response_json = await response.json()
-            return dashboard_display_name in [dashboard.get("displayName") for dashboard in response_json.get("dashboards")]
-    except Exception as e:
-        context.log(f"Failed to list self monitoring dashboards, because: {e}")
-    return False
+    response = await context.gcp_session.request(
+        'GET',
+        url=f"https://monitoring.googleapis.com/v1/projects/dynatrace-gcp-extension/dashboards",
```
karolkieglerski, line 305: why hardcoded project here instead of `{context.project_id_owner}` ?

sq2gxo: good catch! - I was testing locally and forgot to revert

karolkieglerski (approved):

PR #167, opened 2021-09-03
[APM-320834] Cloud Function deployment - replace update.sh and make 

dtwwolinski:
- removed update.sh
- if the function is already deployed there's a new log entry that will be updated

**It's easier to review If you disable the whitespaces**
image


scripts/setup.sh
```diff
@@ -317,9 +317,15 @@ wget -q $FUNCTION_REPOSITORY_RELEASE_URL -O $FUNCTION_ZIP_PACKAGE
 echo "- extracting archive [$FUNCTION_ZIP_PACKAGE]"
 mkdir -p $GCP_FUNCTION_NAME
 unzip -o -q ./$FUNCTION_ZIP_PACKAGE -d ./$GCP_FUNCTION_NAME || exit
-echo "- deploy the function [$GCP_FUNCTION_NAME]"
 pushd ./$GCP_FUNCTION_NAME || exit
 GCP_FUNCTION_TIMEOUT=$(( QUERY_INTERVAL_MIN*60 ))
+
+if [[ $(gcloud functions list --filter=$GCP_FUNCTION_NAME --project="$GCP_PROJECT" --format="value(name)")  ]]; then
+  echo "- updating the function [$GCP_FUNCTION_NAME]"
```
sq2gxo, line 324: I think this is fine, but we should warn user (and ask for confirmation) that any manual modification of function env variables will be overwritten

sq2gxo (dismissed):

scripts/setup.sh
sq2gxo: I function is already deployed we should skip interactive prompts for function size etc...

sq2gxo (commented):
Please add more reviewers

scripts/setup.sh
dtwwolinski: Right now I've changed it in a way that during an update besides the question about the project all other questions are omitted.  Also now no Env variables will be updated as it was in update.sh. Not sure if it's ok, (I'm just worried and looking in the future, how we'll be delivering additional environmental variables in case of additional features), nevertheless it's consistent with the previous approach.

karolkieglerski (dismissed):

scripts/setup.sh
```diff
@@ -317,86 +329,95 @@ wget -q $FUNCTION_REPOSITORY_RELEASE_URL -O $FUNCTION_ZIP_PACKAGE
...
 else
-    GCP_SCHEDULER_CRON="*/${QUERY_INTERVAL_MIN} * * * *"
-    gcloud scheduler jobs create pubsub "$GCP_SCHEDULER_NAME" --topic="$GCP_PUBSUB_TOPIC" --schedule="$GCP_SCHEDULER_CRON" --message-body="x"
+  echo -e "- updating the function \e[1;92m[$GCP_FUNCTION_NAME]\e[0m"
+  gcloud functions -q deploy "$GCP_FUNCTION_NAME" --entry-point=dynatrace_gcp_extension --runtime=python37 --trigger-topic="$GCP_PUBSUB_TOPIC" --service-account="$GCP_SERVICE_ACCOUNT@$GCP_PROJECT.iam.gserviceaccount.com" --ingress-settings=internal-only
```
mswiatkowska, line 340: maybe I'm missing sth but why we don't pass --set-env-vars in case of updating gcp function? I assume that not only new version of gcp function should be installed but also a client would like to start monitoring some new services and we should update env-vars too?

scripts/setup.sh
```diff
@@ -18,7 +18,7 @@ onFailure() {
     exit 2
 }
 
-trap onFailure ERR
+#trap onFailure ERR
```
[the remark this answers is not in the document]
dtwwolinski: This is already fixed.

scripts/setup.sh
dtwwolinski: This is a good question. The functionality is the same as it was in the update.sh file. @sq2gxo Do you think that we should update also the env vars during the update and as you stated before prompt for approval of such action?

sq2gxo: I think we should overwrite env variables - the source of setting should be an activation-config file.
And yes, I think we should ask user for confirmation of update

mswiatkowska: but we need to remember if client choose new services to monitor we also need to upload new dashboards and alerts

scripts/setup.sh
```diff
@@ -317,86 +329,95 @@ wget -q $FUNCTION_REPOSITORY_RELEASE_URL -O $FUNCTION_ZIP_PACKAGE
...
 
-echo -e
-echo "- create self monitoring dashboard"
-SELF_MONITORING_DASHBOARD_NAME=$(cat dashboards/dynatrace-gcp-function_self_monitoring.json | jq .displayName)
-if [[ $(gcloud monitoring dashboards  list --filter=displayName:"$SELF_MONITORING_DASHBOARD_NAME" --format="value(displayName)") ]]; then
-    echo "Dashboard already exists, skipping"
-else
-    gcloud monitoring dashboards create --config-from-file=dashboards/dynatrace-gcp-function_self_monitoring.json
-fi
+if [ "$INSTALL" == true ]; then
```
sq2gxo, line 343: This this part should be run on both update and install, remove the if.
- cloud scheduler should be recreated every time - the queryInterval might have been changed,
- there are checks here for existance of alerts, dashboards.

scripts/setup.sh
```diff
@@ -341,10 +341,20 @@ fi
 
 if [ "$INSTALL" == true ]; then
   echo -e "- deploying the function \e[1;92m[$GCP_FUNCTION_NAME]\e[0m"
+  gcloud functions -q deploy "$GCP_FUNCTION_NAME" --entry-point=dynatrace_gcp_extension --runtime=python37 --memory="$GCP_FUNCTION_MEMORY"  --trigger-topic="$GCP_PUBSUB_TOPIC" --service-account="$GCP_SERVICE_ACCOUNT@$GCP_PROJECT.iam.gserviceaccount.com" --ingress-settings=internal-only --timeout="$GCP_FUNCTION_TIMEOUT" --set-env-vars ^:^GCP_SERVICES=$FUNCTION_GCP_SERVICES:PRINT_METRIC_INGEST_INPUT=$PRINT_METRIC_INGEST_INPUT:DYNATRACE_ACCESS_KEY_SECRET_NAME=$DYNATRACE_ACCESS_KEY_SECRET_NAME:DYNATRACE_URL_SECRET_NAME=$DYNATRACE_URL_SECRET_NAME:REQUIRE_VALID_CERTIFICATE=$REQUIRE_VALID_CERTIFICATE:SERVICE_USAGE_BOOKING=$SERVICE_USAGE_BOOKING:USE_PROXY=$USE_PROXY:HTTP_PROXY=$HTTP_PROXY:HTTPS_PROXY=$HTTPS_PROXY:SELF_MONITORING_ENABLED=$SELF_MONITORING_ENABLED:QUERY_INTERVAL_MIN=$QUERY_INTERVAL_MIN
 else
-  echo -e "- updating the function \e[1;92m[$GCP_FUNCTION_NAME]\e[0m"
+
+  while true; do
+    read -p "- your Environment variables will be updated, do you want to continue? [y/n]" yn
```
sq2gxo, line 348: maybe more descriptive message will be better?
"you Cloud Function will be updated - any manual changes made to Cloud Function environment variables will be replaced with values from 'activation-config.yaml' file, do you want to continue? [y/n]"

dtwwolinski: Your message is more clear, updated.

mswiatkowska (approved):

sq2gxo (approved):

Dawid-Kaszubski (approved):

PR #171, opened 2021-09-21
APM-321122 allow log ingest through Public AG

wojtasskorcz:
Tested using a public AG:
```
deploymentType: "logs"
dynatraceUrl: ""
dynatraceLogIngestUrl: "htts://jxw01498.dev.dynatracelabs.com/"
activeGate.useExisting: "true"
activeGate.dynatracePaasToken: ""
requireValidCertificate: "true"
```
with a Function generating a specific log every minute:
gcp-log-generator-logs
It works!
jxw-gcp-log

### DIFF
--- a/k8s/helm-chart/dynatrace-gcp-function/values.yaml
+++ b/k8s/helm-chart/dynatrace-gcp-function/values.yaml
@@ -33,8 +33,8 @@ dynatraceUrl: ""
 
 # subscription id of log sink pubsub subscription
 logsSubscriptionId: ""
-#dynatraceLogIngestUrl ActiveGate endpoint used to ingest logs to Dynatrace:
-#  - for Public ActiveGate: https://<your_environment_ID>.live.dynatrace.com
+#dynatraceLogIngestUrl endpoint used to ingest logs to Dynatrace:
+#  - for direct ingest through the Cluster API: https://<your_environment_ID>.live.dynatrace.com
 #  - for Environment ActiveGate: https://<active_gate_address>:9999/e/<environment_id> (e.g. https://22.111.98.222:9999/e/abc12345)
 #used ONLY when the activeGate.useExisting is true
 dynatraceLogIngestUrl: ""
--- a/k8s/helm-chart/dynatrace-gcp-function/values.yaml
+++ b/k8s/helm-chart/dynatrace-gcp-function/values.yaml
@@ -33,7 +33,9 @@ dynatraceUrl: ""
 
 # subscription id of log sink pubsub subscription
 logsSubscriptionId: ""
-#dynatraceLogIngestUrl ActiveGate endpoint used to ingest logs to Dynatrace, for example: https://environemnt-active-gate-url:9999/e/environment-id"
+#dynatraceLogIngestUrl ActiveGate endpoint used to ingest logs to Dynatrace:
+#  - for Public ActiveGate: https://<your_environment_ID>.live.dynatrace.com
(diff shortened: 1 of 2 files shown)


k8s/helm-chart/dynatrace-gcp-function/values.yaml
```diff
@@ -33,7 +33,9 @@ dynatraceUrl: ""
 
 # subscription id of log sink pubsub subscription
 logsSubscriptionId: ""
-#dynatraceLogIngestUrl ActiveGate endpoint used to ingest logs to Dynatrace, for example: https://environemnt-active-gate-url:9999/e/environment-id"
+#dynatraceLogIngestUrl ActiveGate endpoint used to ingest logs to Dynatrace:
+#  - for Public ActiveGate: https://<your_environment_ID>.live.dynatrace.com
```
sq2gxo, line 37: 'Public ActiveGate' is our internal term.
rephrase to something like:
"- for direct cluster ingest"

scripts/deploy-helm.sh
```diff
@@ -272,7 +278,10 @@ if [[ $DEPLOYMENT_TYPE == all ]] || [[ $DEPLOYMENT_TYPE == logs ]]; then
   else
     echo "Using an existing Active Gate"
     check_if_parameter_is_empty "$DYNATRACE_LOG_INGEST_URL" "DYNATRACE_LOG_INGEST_URL"
-    check_url "$DYNATRACE_LOG_INGEST_URL" "$ACTIVE_GATE_TARGET_URL_REGEX" "Not correct dynatraceLogIngestUrl. Example of proper ActiveGate endpoint used to ingest logs to Dynatrace: https://<your_activegate_IP_or_hostname>:9999/e/<your_environment_ID>"
+    check_url "$DYNATRACE_LOG_INGEST_URL" "$DYNATRACE_URL_REGEX" "$ACTIVE_GATE_TARGET_URL_REGEX" \
+      "Not correct dynatraceLogIngestUrl. Example of proper ActiveGate endpoint used to ingest logs to Dynatrace:\n
+        - for Public ActiveGate: https://<your_environment_ID>.live.dynatrace.com\n
```
sq2gxo, line 283: change message - same as in comment above

sq2gxo: our DYNATRACE_URL_REGEX is already supporting managed (we already have one customer), provbide example in message

k8s/helm-chart/dynatrace-gcp-function/values.yaml
wojtasskorcz: You're right! I only learned about it today. We have a discussion ongoing with Teo and Michał F. on how to call it so that it's also consistent with our official docs

scripts/deploy-helm.sh
wojtasskorcz: I'm not sure we should mention it explicitly in docs or script help messages if it's in preview phase. Or would you like to have a disclaimer like:
```
- for direct ingest through the SaaS Cluster API: https://<your_environment_ID>.live.dynatrace.com\n
- for direct ingest through the Managed Cluster API (preview, only selected customers): https://<your_environment_ID>.live.dynatrace.com\n
```

sq2gxo: You are right, this is a preview, please disregard my previous comment about managed

wojtasskorcz: Alright, adjusted the message. Please re-review

sq2gxo (approved):

PR #176, opened 2021-09-28
[APM-326992] GCP- project env

dtwwolinski:
- [k8s installation] project_id is taken now from environment variable GCP_PROJECT or in case it's not available(tests) a default "dynatrace-gcp-extension" value is used
- added documentation.

### DIFF
--- a/src/lib/credentials.py
+++ b/src/lib/credentials.py
@@ -94,7 +94,7 @@ async def create_default_service_account_token(context: LoggingContext, session:
 
 
 def get_project_id_from_environment():
-    return os.environ.get("GCP_PROJECT", "dynatrace-gcp-extension")
+    return os.environ.get("GCP_PROJECT")
 
 
 async def create_token(context: LoggingContext, session: ClientSession):
--- a/src/lib/credentials.py
+++ b/src/lib/credentials.py
@@ -94,7 +94,7 @@ async def create_default_service_account_token(context: LoggingContext, session:
 
 
 def get_project_id_from_environment():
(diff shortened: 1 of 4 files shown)


src/lib/credentials.py
```diff
@@ -94,7 +94,7 @@ async def create_default_service_account_token(context: LoggingContext, session:
 
 
 def get_project_id_from_environment():
-    return os.environ.get("GCP_PROJECT")
+    return os.environ.get("GCP_PROJECT", "dynatrace-gcp-extension")
```
sq2gxo, line 97: It is fixing problem, but in case of misconfiguration we will silently switch to default value.
If "dynatrace-gcp-extension" value is needed for tests only - I'd rather mock it in tests, same way as other env variables.
Then in main.py you could throw exception if GCPPROJECT is not defined, same as in https://github.com/dynatrace-oss/dynatrace-gcp-function/blob/master/src/lib/logs/log_forwarder.py#L57

dtwwolinski: It's not an issue to mock this value at the tests and **I'll change** it since it's more readable that way. Nevertheless, to keep it clear, a little explanation. The default value will be only used in tests. For GCP function installation the GCP_PROJECT environment variable is available by default. For k8s installation, the environment variable is also available since it's initiated by the helm chart function-deployment.yml.

dtwwolinski: Done

mswiatkowska (approved):

sq2gxo (approved):

Dawid-Kaszubski (approved):

PR #179, opened 2021-09-30
Apm 327142 repackage gcp function

mswiatkowska:


karolkieglerski (approved):

scripts/setup.sh
```diff
@@ -340,6 +341,34 @@ else
...
+cd ../extensions || exit
+echo "- choosing extensions to upload to Dynatrace"
+for EXTENSION_ZIP in *.zip; do
+  EXTENSION_NAME="$(basename "$EXTENSION_ZIP" .zip)"
+  unzip -j -q "$EXTENSION_ZIP" "extension.zip"
+  unzip -p -q "extension.zip" "extension.yaml" > "$EXTENSION_NAME".yaml
+  EXTENSION_GCP_CONFIG=$(yq e '.gcp' "$EXTENSION_NAME".yaml)
+  SERVICES=$(echo "$EXTENSION_GCP_CONFIG" | yq e -j | jq -r 'to_entries[] | "\(.value.service)/\(.value.featureSet)"')
+  REMOVE_EXTENSION=true
+  for SERVICE in $SERVICES; do
```
sq2gxo, line 353: we also need to support case when service is configured without /fetureSet added, e.g. 'cloud_function' should be treated as a 'cloud_function/default'.
See https://github.com/dynatrace-oss/dynatrace-gcp-function/blob/master/src/main.py#L90

scripts/setup.sh
```diff
@@ -97,6 +97,7 @@ readonly DYNATRACE_URL_SECRET_NAME=$(yq e '.googleCloud.common.dynatraceUrlSecre
 readonly DYNATRACE_ACCESS_KEY_SECRET_NAME=$(yq e '.googleCloud.common.dynatraceAccessKeySecretName' $FUNCTION_ACTIVATION_CONFIG)
 readonly FUNCTION_GCP_SERVICES=$(yq e -j '.activation.metrics.services' $FUNCTION_ACTIVATION_CONFIG | jq 'join(",")')
 readonly SERVICES_TO_ACTIVATE=$(yq e -j '.activation.metrics.services' $FUNCTION_ACTIVATION_CONFIG | jq -r .[] | sed 's/\/.*$//')
+SERVICES_FROM_ACTIVATION_CONFIG=($(yq e -j '.activation.metrics.services' $FUNCTION_ACTIVATION_CONFIG | jq -r .[] ))
```
karolkieglerski, line 100: can't be `readonly`?

scripts/setup.sh
```diff
@@ -340,6 +341,66 @@ else
...
+        echo "$EXTENSION_GCP_CONFIG" >> ../dynatrace-gcp-function/config/"${BASH_REMATCH[1]}".yaml
+      fi
+      break
+    fi
+  done
+  rm extension.zip
+  rm "$EXTENSION_NAME".yaml
+  if $REMOVE_EXTENSION; then rm "$EXTENSION_ZIP"; fi
+done
+rm -r ../extensions_from_cluster
```
karolkieglerski, line 400: use `-f` because it will not be empty then will be an error

scripts/setup.sh
```diff
@@ -340,6 +341,66 @@ else
...
 fi
 
+# If --upgrade option is not set, all gcp extensions are downloaded from the cluster to get configuration of gcp services for version that is currently active on the cluster.
+if [[ "$UPGRADE_EXTENSIONS" != "Y" ]]; then
+  echo "- downloading active extensions from Dynatrace"
+  mkdir -p ../extensions_from_cluster
+  cd ../extensions_from_cluster || exit
+  EXTENSIONS_TO_DOWNLOAD="com.dynatrace.extension.google"
+  readarray -t EXTENSIONS_FROM_CLUSTER_ARRAY <<<"$( echo "${EXTENSIONS_FROM_CLUSTER}" | sed 's/ /\n/' | grep "$EXTENSIONS_TO_DOWNLOAD" )"
+  for i in "${!EXTENSIONS_FROM_CLUSTER_ARRAY[@]}"; do
```
karolkieglerski, line 351: use different name then `i`

scripts/setup.sh
```diff
@@ -340,6 +341,66 @@ else
...
+    EXTENSION_NAME="$(cut -d':' -f1 <<<"${EXTENSIONS_FROM_CLUSTER_ARRAY[$i]}")"
+    EXTENSION_VERSION="$(cut -d':' -f2 <<<"${EXTENSIONS_FROM_CLUSTER_ARRAY[$i]}")"
+    curl -s -X GET "${DYNATRACE_URL}/api/v2/extensions/${EXTENSION_NAME}/${EXTENSION_VERSION}" -H "Accept: application/octet-stream" -H "Authorization: Api-Token ${API_TOKEN}" -o "${EXTENSION_NAME}-${EXTENSION_VERSION}.zip"
```
karolkieglerski, line 354: add `-k` because it can self signed certificate

scripts/setup.sh
```diff
@@ -340,6 +341,66 @@ else
...
+  done
+
+  for EXTENSION_FROM_CLUSTER_ZIP in *.zip; do
```
karolkieglerski, line 357: this for is not needed, you can move it to above for

scripts/setup.sh
mswiatkowska: I left it as is because it's the variable for index

scripts/setup.sh
mswiatkowska: sorry that was before committing review changes, I was talking about this line: `SERVICES_FROM_ACTIVATION_CONFIG[$i]="${SERVICES_FROM_ACTIVATION_CONFIG[$i]}/default"`
if it is `readonly` I'm getting `SERVICES_FROM_ACTIVATION_CONFIG: readonly variable`

PR #183, opened 2021-10-05
Apm 327665 handle schema changes in python code

mswiatkowska:


tests/testresources/metrics/config/api.yml
```diff
@@ -9,47 +9,47 @@ gcp:
...
+      samplePeriod: 60
       valueType: INT64
       metricKind: DELTA
+      unit: '1'
     dimensions:
     - value: metric.labels.protocol
     - value: metric.labels.response_code
     - value: metric.labels.response_code_class
     - value: metric.labels.grpc_status_code
   - value: serviceruntime.googleapis.com/api/request_latencies
```
sq2gxo, line 43: can we have prefixes 'metric:' and 'label:' added in some of test configs?
I know they are optional, but it is worth to have it tested.

sq2gxo (approved):

PR #185, opened 2021-10-11
Feature/apm 320851 gcp function k8s fetch services config from dynatrace during startup

Dawid-Kaszubski:
Added fetching services configuration contained in extensions from Dynatrace cluster. 
Removed uploading alerts and dashboards - it'll be covered in extension flow.


Dawid-Kaszubski:
> https://github.com/dynatrace-oss/dynatrace-gcp-function/blob/master/src/lib/metrics.py#L104 line is failing and https://github.com/dynatrace-oss/dynatrace-gcp-function/blob/master/src/lib/metrics.py#L111 one also. Please adjust this PR to the new schema 1.229

Checked with the current version. Working correctly.

PR #203, opened 2021-11-02
Change image tag to v2

sq2gxo:


build/docker-deploy.sh
```diff
@@ -11,9 +11,9 @@ if [[ "${PUSH:-}" == "true" ]]; then
     base64 -d >~/.docker/config.json <<<"$OAO_DOCKER_AUTH"
 
     ./build/version.sh
-    docker tag dynatrace/dynatrace-gcp-function:latest dynatrace/dynatrace-gcp-function:$TAG
+    docker tag dynatrace/dynatrace-gcp-function:v2-latest dynatrace/dynatrace-gcp-function:$TAG
```
karolkieglerski, line 14: consider to not change it to `v1-latest` because we are moving from `0.x` to `1.x`

dtwwolinski: Shouldn't it be more like ```dynatrace/dynatrace-gcp-function/v1:latest``` ?

sq2gxo: No, image name should not change. It's just the tag change from "latest" to v1-latest"

karolkieglerski (approved):

sq2gxo:
Confirmed that both tags (v1-latest and tag) should work correctly by pushing image with two tags to gcr.io
image

build/docker-deploy.sh
```diff
@@ -11,9 +11,9 @@ if [[ "${PUSH:-}" == "true" ]]; then
     base64 -d >~/.docker/config.json <<<"$OAO_DOCKER_AUTH"
 
     ./build/version.sh
```
karolkieglerski, line 13: please move it the same as on master

PR #209, opened 2021-11-04
Feature/apm 332220 change validation in installations scripts to accept only services with featuresets

Dawid-Kaszubski:


karolkieglerski:
looks ok to me, but please squash commits before merging it :)

karolkieglerski (approved):

scripts/setup.sh
```diff
@@ -572,12 +567,10 @@ if [[ "$UPGRADE_EXTENSIONS" != "Y" && -n "$EXTENSIONS_FROM_CLUSTER" ]]; then
   done
 fi
 
-# Add '/default' to service name when featureSet is missing
```
mswiatkowska, line 575: Why did you remove it? Shouldn't we change `default` to `default_metrics` here?
Line 586 should also be change anyway.

Dawid-Kaszubski: We don't want to add any featureSet by default. FeatureSet for service should be defined explicitly.

Dawid-Kaszubski: We don't want to add any featureSet by default. FeatureSet for service should be defined explicitly.

PR #215, opened 2021-11-08
Apm 334309 update api token requirements

sq2gxo:


karolkieglerski:
@sq2gxo why? this is required

sq2gxo:
> @sq2gxo why? this is required https://www.dynatrace.com/support/help/extend-dynatrace/extensions20/extension-lifecycle/#permissions

You are right, I've tried install withot Read/WriteConfig permission and it worked (probably due to fact that we are providing assets only) but it is safer to keep the requirement.

PR #223, opened 2021-11-10
APM-335445 wait helm untill deploy

karolkieglerski:


gcp_iam_roles/dynatrace-gcp-function-deployment-role.yaml
```diff
@@ -4,26 +4,24 @@ title: Dynatrace GCP Function deployment role
...
-  - pubsub.topics.create
-  - pubsub.topics.update
-  - pubsub.topics.getIamPolicy
-  - pubsub.topics.setIamPolicy
-  - pubsub.topics.attachSubscription
   - container.clusters.get
   - container.configMaps.create
   - container.configMaps.delete
   - container.configMaps.get
+  - container.configMaps.update
```
sq2gxo, line 11: I don't understand why those changes are her? This PR is about "helm wait"?

karolkieglerski: helm wait needs sample of additional requirements to check state of each components ex. replicasets that are not added before. This file is damp of our role that is working right now (this is not added automatically)

dtwwolinski: Is this written somewhere?

mswiatkowska: but `container.configMaps.update` permission is needed in case when some rubbish are left from previous e2e execution - maybe after changing delete to uninstall (on master branch) this permission is not needed?

But I'm also wondering - is this permission needed for the client to run upgrade helm deployment?

karolkieglerski: to do upgrade is needed

karolkieglerski: all update is not needed if we install fresh new each time

mswiatkowska (approved):

dtwwolinski (approved):

sq2gxo (approved):

PR #227, opened 2021-11-12
Use helm uninstall instead of delete

mswiatkowska:


sq2gxo (approved):

tests/e2e/before-deployment-test-cleanup.sh
```diff
@@ -13,6 +13,8 @@
 #     See the License for the specific language governing permissions and
 #     limitations under the License.
 
+helm -n dynatrace ls --all --short | grep dynatrace-gcp-function | xargs -L1 helm -n dynatrace uninstall
```
dtwwolinski, line 16: Please add --wait 
https://helm.sh/docs/helm/helm_uninstall/

dtwwolinski (changes requested):

dtwwolinski (approved):

tests/e2e/before-deployment-test-cleanup.sh
mswiatkowska: It occurred that --wait is not supported by uninstall command.

Dawid-Kaszubski (approved):

PR #237, opened 2021-11-30
APM-336055 deployment after each new release

karolkieglerski:


Dawid-Kaszubski (approved):

scripts/deploy-function.sh
```diff
@@ -0,0 +1,190 @@
...
+    iamRole: "${IAM_ROLE_METRCICS}"
+    requireValidCertificate: false
+  metrics:
+    pubSubTopic: "${PUBSUB_TOPIC}"
+    function: "${METRIC_FORWARDING_FUNCTION}"
+    scheduler: "${CLOUD_SCHEDULER}"
+activation:
+  metrics:
+    services:
+    # Google Cloud APIs
```
mswiatkowska, line 38: should be changed to the new config: https://github.com/dynatrace-oss/dynatrace-gcp-function/blob/APM-310059-move-gcp-assets-to-ef/activation-config.yaml#L48

mswiatkowska (changes requested):

.travis.yml
```diff
@@ -105,87 +100,87 @@ jobs:
...
-  #     - ./tests/e2e/execute-tests.sh
-  #   after_script:
-  #     - ./tests/e2e/deployment-test-cleanup.sh
+  - stage: e2e-helm-deployment-test
+    name: E2E GCP helm deployment test
+    sudo: required
+    language: python
+    python:
+      - "3.8"
+    if: (branch = master OR type = pull_request) AND fork = false
```
kamiljsokolowski: What is the deployment strategy here? Test the deployment on each push to `master` or on each new release?

If you plan to run deployment tests on each push to `master` or when a PR is created, then the PR title ("deploy after each new release") is somewhat misleading.

.travis.yml
```diff
@@ -105,87 +100,87 @@ jobs:
...
+      - PUBSUB_SUBSCRIPTION="e2e_test_subscription_${TRAVIS_BUILD_ID}"
+      - IAM_SERVICE_ACCOUNT="e2e-test-sa-${TRAVIS_BUILD_ID}"
+      - LOG_ROUTER="e2e_test_log_router_${TRAVIS_BUILD_ID}"
+      - IAM_ROLE_PREFIX="e2e_test_${TRAVIS_BUILD_ID}"
+      - K8S_CLUSTER="e2e-test-master"
+      - GCP_PROJECT_ID="dynatrace-gcp-extension"
+      - GCR_NAME="gcr.io/${GCP_PROJECT_ID}/dynatrace-gcp-function"
+      - E2E=true
+      - CLOUD_FUNCTION_NAME="e2e_test_sample_app"
+      - secure: "muHGNxIK6fxTltIsGgms6v3dleuQ1LZyQmZIGQDSfQ5sfYDYEvR2xsabK237AipSSGppoVhc7S6ri8A/WNQautHUfdBOsVn1i7czzoPhELdOJkAiigXrI3kla68OTCiC0TPwWbIYOxzmiYDHZr+rYbA1773O0iyhWHiTVQerLMndn0wCFWUUnLbD1WfzdAxlfSo1rd7u81m8g4efLIUppo/fjfaxmVwvo1oIfensGVknyOE+gtSh/OpGdYfNeOqxoDGOoMe5bnCqibUoREXRzUZKGb6upxMMmW6E3hGG02kqle4jYjaC7Uiu2IVZsK38iAWNGwIRy21CIXNHn4R6we44n5JLH92aiRtVw8D/fDOnN2gWTtmdrvoB8YklbaYZBnDrGJxI6rESfeQjwD8ImWWKUJFUlwovKjNWhOw8RMhPVtPOuC00+F2UDq5NosrWYGUjLjaKFC/4nysOiK2DgsMBF/Mk1Ei1BDMYmcHQ7Jsrs8ys/kgPasV3Qs9xOuSJpF+AiwKPfI1R9bxYiORg70TNz62wPD7FlKWIVe0DvhrWcHCTpnrROabt6B0hUROfljCZe2H+j8HD9uCJCJHk5qP7llNVofVnz8w+s90wLEKpUNn2rVUH0fzffYM7YaLRPhzLEUJSQso3a7hXFfcoUow8OOPTmCgViGCfPjuy2zc="
```
kamiljsokolowski: I can see that those env variables are being reused in several stages. Wouldn't it be easier to store them under a single variable and call that variable instead?

karolkieglerski: it duplicates i 2 sections but in one section this is different

.travis.yml
karolkieglerski: this is `e2e-helm-deployment-test` section

.travis.yml
```diff
@@ -61,6 +64,42 @@ jobs:
...
       on:
         tags: true
+
+  - stage: github-install-release
+    name: GitHub install release
+    sudo: required
+    language: python
+    python:
+      - "3.8"
+    if: tag IS present
```
kamiljsokolowski, line 74: Why not `tag =~ /^release-.*` here as in `docker-deploy` stage?

.travis.yml
kamiljsokolowski: As it turns out, this comment was pointing to the pre-PR version of the code. Either my mistake or GH is playing hide-and-seek.

.travis.yml
kamiljsokolowski: If those tokens are (or will be) subjects to change it might be worth moving them to the top of the pipeline and assign to env vars after all. This would make the code more readable... and DRY gods satisfied;)

karolkieglerski: only one `- secure: "tdw5iV7XGhQ+gVPw70n+/SeA2xfVAXL5uwhMrKjI6F****` is in 2 stages but not needed on all stages so I don't see any advantages to move it to global

mswiatkowska (approved):

PR #245, opened 2021-12-08
APM-340751: Fixed dashboards upload for k8s deployment

mswiatkowska:


src/lib/utilities.py
```diff
@@ -11,9 +11,38 @@
...
+def read_activation_yaml():
+    activation_file_path = '/code/config/activation/gcp_services.yaml'
+    try:
+        with open(activation_file_path, encoding="utf-8") as activation_file:
+            activation_yaml = yaml.safe_load(activation_file)
+    except Exception:
+        activation_yaml = yaml.safe_load(os.environ.get("ACTIVATION_CONFIG", ""))
+    if not activation_yaml:
+        activation_yaml = {}
+    return activation_yaml
```
karolkieglerski, line 36: If you allow return `{}` how `load_activated_feature_sets` will work? maybe throw exception here or in `load_activated_feature_sets`

mswiatkowska: Thanks for catching that!
I forgot to add empty list as default value...

karolkieglerski (approved):

Dawid-Kaszubski (approved):

PR #248, opened 2021-12-10
Fix correct managed cluster parser

karolkieglerski:


k8s/helm-chart/dynatrace-gcp-function/templates/_helpers.tpl
```diff
@@ -21,7 +21,7 @@
     {{- if contains "/e/" .dynatraceUrl -}}
       {{- printf "%s" (split "/e/" .dynatraceUrl)._1 | replace "/" "" }}
```
mswiatkowska, line 22: this case is for managed url

wojtasskorcz (approved):

mswiatkowska (approved):

PR #252, opened 2021-12-14
Extensions fixes

mswiatkowska:


scripts/setup.sh
```diff
@@ -387,7 +387,7 @@ fi
 
 echo -e
 echo "- schedule the runs"
-if [[ $(gcloud scheduler jobs list --filter=name:$GCP_SCHEDULER_NAME --format="value(name)") ]]; then
+if [[ $(gcloud scheduler jobs list --filter="name ~ $GCP_SCHEDULER_NAME$" --format="value(name)") ]]; then
```
karolkieglerski, line 390: `$` should be here?

karolkieglerski (approved):

scripts/setup.sh
mswiatkowska: yes, actually it's regex pattern, the format with ":" is deprecated for scheduler jobs list command

mswiatkowska:
```
WARNING: --filter : operator evaluation is changing for consistency across Google APIs. 
name:dynatrace-gcp-schedule currently matches but will not match in the near future.
```

Dawid-Kaszubski (approved):

karolkieglerski (approved):

PR #260, opened 2021-12-16
double check gcp project ID

karolkieglerski:


mswiatkowska (approved):

scripts/deploy-helm.sh
```diff
@@ -172,6 +172,10 @@ check_s3_url
 
 if [ -z "$GCP_PROJECT" ]; then
   GCP_PROJECT=$(gcloud config get-value project 2>/dev/null)
+  if [ -z "$GCP_PROJECT" ]; then
+  err "Invalid GCP_PROJECT: $GCP_PROJECT. Set correct gcpProjectId in values.yaml"
```
mswiatkowska, line 176: or we could just write that it's not set (it's always be empty in this place)

mswiatkowska (approved):

PR #266, opened 2022-01-05
Remove pip install suggestion. In cloudshell it install by default 2.…

mmajcher:
…X version which is not compatible

### DIFF
--- a/scripts/lib.sh
+++ b/scripts/lib.sh
@@ -65,14 +65,13 @@ versionNumber() {
 
 test_req_yq() {
   if ! command -v yq &>/dev/null; then
-    err 'yq and jq is required to install Dynatrace function. Please refer to following links for installation instructions:
+    err 'yq (4.9.x) and jq is required to install Dynatrace function. Please refer to following links for installation instructions:
       YQ: https://github.com/mikefarah/yq
       Example command to install yq:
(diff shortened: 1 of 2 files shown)


wojtasskorcz:
I'd also add in line 34 (cannot comment inline there) which version we specifically require: "yq in 4+version". Right now, if someone doesn't have yq, they may install an unsupported version from their package manager instead of the official page and end up having to reinstall. This may be a frustrating experience

scripts/lib.sh
```diff
@@ -65,14 +65,13 @@ versionNumber() {
 
 test_req_yq() {
   if ! command -v yq &>/dev/null; then
-    err 'yq and jq is required to install Dynatrace function. Please refer to following links for installation instructions:
+    err 'yq (4.9.x) and jq is required to install Dynatrace function. Please refer to following links for installation instructions:
```
wojtasskorcz, line 68: Don't want to stall you with this PR so I'll approve. Just wanted to mention that here we're saying 4.9.x. In the condition in line 85 we're checking for >=4.9.8. In the error message in line 86 we're saying 4+. It's all a bit chaotic. Please unify this if you wish, but no pressure from my side

wojtasskorcz (approved):

scripts/lib.sh
Dawid-Kaszubski: Fully agree that became chaotic. Scripts require yq in version 4.9.8 or higher.

karolkieglerski (approved):

PR #267, opened 2022-01-10
APM-339646 init full log to file

karolkieglerski:


wojtasskorcz:
This is pretty bad for script readability. Is there not global flag in bash or sth that would redirect output both to stdout and to a file (given once, not every time)? Maybe creating a "wrapper script" that would just call
```
setup.sh <args> | tee -a "$FULL_LOG_FILE"
```
?

mmajcher:
+1, I would expect at least a wrapper (sounds most simple)

mmajcher:
Maybe you can enclose all "top level logic" into some function and then execute function with tee? 
Also I think stderr would be helpful  https://stackoverflow.com/a/418899

karolkieglerski:
> This is pretty bad for script readability and future maintenance (we'd have to remember about adding this tee suffix every time). Is there no global flag in bash or sth that would redirect output both to stdout and to a file (given once, not every time)? Maybe creating a "wrapper script" that would just call
> 
> ```
> setup.sh <args> | tee -a "$FULL_LOG_FILE"
> ```
> 
> ?

and how you will deliver that script? what about security data like tokens? what if customer will run setup script it selves? want with ask parameters?

karolkieglerski:
@mmajcher and @wojtasskorcz please first read changes then ask questions or ideas... tee is not added to all please check. it a lot of reasons why

karolkieglerski:
plus we want to have more in log file instead of console. please diff files in jira task

mmajcher:
Delivery is not a problem.
Approach 1 - have a main function and run it inside setup.sh like this: "main $@ | tee XXXX" 
Approach 2 if 1 does not work - have separate file setup.sh which will run setup-internal.sh with tee and both are inside our package.

Basic approach would be to have exactly what customer sees. Any additional outputs (like curl debug etc.) would be fine if are easy to achieve. 

Only issue I see - do we print any tokens that would be caught with this approach? Can you double check that? I don't think there is any need at all to print any token to stdout.

scripts/deploy-helm.sh
```diff
@@ -234,7 +234,7 @@ if [[ $DEPLOYMENT_TYPE == all ]] || [[ $DEPLOYMENT_TYPE == logs ]]; then
   if [[ $USE_EXISTING_ACTIVE_GATE == false ]]; then
     check_dynatrace_docker_login
   else
-    echo "Using an existing Active Gate"
+    echo "Using an existing Active Gate" | tee -a "$FULL_LOG_FILE"
     check_if_parameter_is_empty "$DYNATRACE_LOG_INGEST_URL" "DYNATRACE_LOG_INGEST_URL"
     check_url "$DYNATRACE_LOG_INGEST_URL" "$DYNATRACE_URL_REGEX" "$ACTIVE_GATE_TARGET_URL_REGEX" \
```
wojtasskorcz, line 239: I'm looking at examples where `tee` is not used, as you advised. Here we're not using it neither at call site nor in the `check_url` implementation. What's the reason?

karolkieglerski: because in `check_url` function we are using `err` function to return message and there `tee` is added

karolkieglerski: this is better example when we want to have information in file but on in stdout https://github.com/dynatrace-oss/dynatrace-gcp-function/pull/267/files#diff-5838719667e50a69f27dfddac131cad8ed4e11bee597a64a1b48454efdfad40bR217

wojtasskorcz:
After our discussion at today's daily where you said that this PR was pretty quick to prepare, I'm a bit less worried about the amount of potential future work (in context of maintenance). I recognize the usefulness of all the additional info that we can log in the file with the current approach and which we'd have to omit if we just redirected user output to file. I could approve it on this premise, but would still like to hear @kamiljsokolowski and maybe @dtmawo share their thoughts on this. Guys?

kamiljsokolowski (approved):

Dawid-Kaszubski (dismissed):

wojtasskorcz:
Great idea to introduce the `debug` and `info` functions!

wojtasskorcz (dismissed):

wojtasskorcz (approved):

wojtasskorcz (approved):

mmajcher (approved):

PR #269, opened 2022-01-11
[APM-345587] Changed `activeGate.useExisting` default value to true i…

Dawid-Kaszubski:
Changed `activeGate.useExisting` default value to true in values.yaml.
Since the public ingest for logs and metrics is available, using an existing Active Gate should be used as a default approach.

### DIFF
--- a/k8s/helm-chart/dynatrace-gcp-function/values.yaml
+++ b/k8s/helm-chart/dynatrace-gcp-function/values.yaml
@@ -42,8 +42,8 @@ logsSubscriptionId: ""
 dynatraceLogIngestUrl: ""
 
 activeGate:
-  #useExisting if set to false (by default) the Active Gate will be deployed in k8s. In that case also the activeGate.dynatracePaasToken must be filled.
-  useExisting: "false"
+  #useExisting if set to true (by default) integration will used existing Active Gate, if it set to false the Active Gate will be deployed in k8s. In that case also the activeGate.dynatracePaasToken must be filled.
+  useExisting: "true"
   #dynatracePaasToken PaaS token, generated in Dynatrace cluster.
   dynatracePaasToken: ""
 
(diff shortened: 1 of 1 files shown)


karolkieglerski (approved):

k8s/helm-chart/dynatrace-gcp-function/values.yaml
```diff
@@ -42,8 +42,8 @@ logsSubscriptionId: ""
 dynatraceLogIngestUrl: ""
 
 activeGate:
-  #useExisting if set to false (by default) the Active Gate will be deployed in k8s. In that case also the activeGate.dynatracePaasToken must be filled.
-  useExisting: "false"
+  #useExisting if set to true (by default) integration will used existing Active Gate, if it set to false the Active Gate will be deployed in k8s. In that case also the activeGate.dynatracePaasToken must be filled.
```
mswiatkowska, line 45: imo the message should be similar to the one for Azure/AWS log forwarder (it's important to mention about ingest through the Cluster API):
`If you choose to use direct ingest through the Cluster API or existing ActiveGate, put 'true'.
If you choose new ActiveGate deployment, put 'false'. In such case, ActiveGate will be deployed in k8s...`

Dawid-Kaszubski: Good idea, thanks!

mswiatkowska (approved):

k8s/helm-chart/dynatrace-gcp-function/values.yaml
```diff
@@ -42,7 +42,8 @@ logsSubscriptionId: ""
 dynatraceLogIngestUrl: ""
 
 activeGate:
-  #useExisting if set to true (by default) integration will used existing Active Gate, if it set to false the Active Gate will be deployed in k8s. In that case also the activeGate.dynatracePaasToken must be filled.
+  #useExisting if you choose to use direct ingest through the Cluster API or existing ActiveGate, put 'true' (default).
```
mmajcher, line 45: "the Cluster API (available to all SaaS customers)"

mmajcher (approved):

PR #270, opened 2022-01-11
Explain IAM role update - probably you can safely replace it

mmajcher:


mmajcher:
Trying to address this 
image

mmajcher:
Maybe we should add this etag thing? Because this message does look scary indeed @Dawid-Kaszubski

Dawid-Kaszubski:
> Maybe we should add this etag thing? Because this message does look scary indeed
Sounds good, but it'll require more work, than just adding etag to gcloud command.
It requires additional `iam.roles.describe` permission for deployment role and modification of described role in the script.
https://cloud.google.com/iam/docs/creating-custom-roles#editing_an_existing_custom_role
IMO great candidate for improvement task.

karolkieglerski (approved):

wojtasskorcz (approved):

PR #271, opened 2022-01-14
bugfix/APM-344818

kamiljsokolowski:
After some discussion with @mswiatkowska we decided not to complicate the setup and to remove the option to auto-set the GCP Project ID based on cloud or local shell settings. Based on the docs, users are required to set the parameter in `values.yaml` file before script execution and should receive an error message otherwise.

### DIFF
--- a/scripts/deploy-helm.sh
+++ b/scripts/deploy-helm.sh
@@ -170,10 +170,7 @@ API_TOKEN_SCOPES=('"logs.ingest"' '"metrics.ingest"' '"ReadConfig"' '"WriteConfi
 
 check_s3_url
 
-if [ -z "$GCP_PROJECT" ]; then
-  err "Invalid GCP_PROJECT. Set correct gcpProjectId in values.yaml"
-  exit 1
-fi
+check_if_parameter_is_empty "$GCP_PROJECT" "Set correct gcpProjectId in values.yaml"
 
 gcloud config set project "$GCP_PROJECT"
 echo "- Deploying dynatrace-gcp-function in [$GCP_PROJECT]"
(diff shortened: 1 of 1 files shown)


scripts/deploy-helm.sh
```diff
@@ -171,11 +171,8 @@ API_TOKEN_SCOPES=('"logs.ingest"' '"metrics.ingest"' '"ReadConfig"' '"WriteConfi
 check_s3_url
 
 if [ -z "$GCP_PROJECT" ]; then
-  GCP_PROJECT=$(gcloud config get-value project 2>/dev/null)
-  if [ -z "$GCP_PROJECT" ]; then
-    err "Invalid GCP_PROJECT. Set correct gcpProjectId in values.yaml"
-    exit 1
-  fi
+  err "Invalid GCP_PROJECT. Set correct gcpProjectId in values.yaml"
```
mswiatkowska, line 174: (minor) Kamil, sorry I didn't mention earlier, we have `check_if_parameter_is_empty` function in lib,sh and we could use it here.

kamiljsokolowski: Ok. The question is do we have to use it or just keep the fix as is?

mswiatkowska: I'd prefer to change it especially that message is about missing required parameter which is a bit more adequate than `Invalid GCP_PROJECT` and we will reuse our existing function.

kamiljsokolowski: Updated.

mswiatkowska (approved):

karolkieglerski (approved):

wojtasskorcz (approved):

PR #274, opened 2022-01-25
Embedded ext tools

dtmawo:
Add required external binaries for supported architecture.


build/get_ext_tools.sh
```diff
@@ -0,0 +1,23 @@
+EXT_DIR=ext_tools
```
karolkieglerski, line 1: this is not correct bash file missing `#!/usr/bin/env bash`

karolkieglerski:
please remove binaries from git with rewriting history to not increasing repository size but `build/get_ext_tools.sh` run at build stage as name of this catalog tells

scripts/uninstall.sh
```diff
@@ -13,18 +13,11 @@
...
 #     limitations under the License.
 
-err() {
-  MESSAGE=$1
-  echo -e >&2
-  echo -e "\e[91mERROR: \e[37m$MESSAGE" >&2
-  echo -e >&2
-}
-
-versionNumber() {
```
karolkieglerski, line 23: why you removed this? this function don't exist in lib

scripts/uninstall.sh
```diff
@@ -33,34 +26,6 @@ TMP_FUNCTION_DIR=$(mktemp -d)
 echo -e "\033[1;34mDynatrace function for Google Cloud Platform monitoring / uninstall script"
 echo -e "\033[0;37m"
 
-if ! command -v yq &> /dev/null
```
karolkieglerski, line 36: it is also required to check here if yq and jq exist

tests/e2e/deployment-test-cloud-function.sh
```diff
@@ -19,24 +19,23 @@ check_function_state()
...
 # Run cloud function deployment.
 rm -rf ./e2e_test
-mkdir -p ./e2e_test/gcp_iam_roles
-cp ./scripts/lib.sh ./e2e_test/lib.sh
-cp ./scripts/setup.sh ./e2e_test/setup.sh
-cp dynatrace-gcp-function.zip ./e2e_test/dynatrace-gcp-function.zip
-cp ./gcp_iam_roles/dynatrace-gcp-function-metrics-role.yaml ./e2e_test/gcp_iam_roles/
+mkdir ./e2e_test
+unzip -d ./e2e_test ./artefacts/function-deployment-package.zip
+cp ./artefacts/dynatrace-gcp-function.zip ./e2e_test
```
karolkieglerski, line 36: why you extracting to e2e_test and then copy this zip to the same directory?

mswiatkowska:
Do we have jira ticket for removing jq/yq requirements from doc?

scripts/uninstall.sh
dtmawo: init_ext_tools does this check

tests/e2e/deployment-test-cloud-function.sh
dtmawo: function-deployment-package.zip doesn't contain dynatrace-gcp-function.zip. e2e tests assumes local zip file (git release is not available for download)

scripts/uninstall.sh
dtmawo: It's there: https://github.com/dynatrace-oss/dynatrace-gcp-function/blob/embedded_ext_tools/scripts/lib.sh#L62

dtmawo:
Not yet. Anyway, except of the uninstall.sh, old flow is supposed to be compatible with the new package, just downloading yq/jq is not needed, so I think there is no need to strictly synchronize docs with release.

build/get_ext_tools.sh
dtmawo: Fixed

karolkieglerski:
Screenshot 2021-12-08 at 10 58 54
binary changes still in diff

karolkieglerski:
Screenshot 2022-01-27 at 10 52 14
binary still in diff

build/get_ext_tools.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
```
mswiatkowska, line 1: What is the reason of having this file? It's not used, right?

build/get_ext_tools.sh
karolkieglerski: I was suggesting better solution
> Run the program through /usr/bin/env, the user does not need to find the location of the program in the system (because the location of the command or program may be different in different systems), as long as the program is in your $PATH;
>
>Another advantage of running the program through /usr/bin/env is that it will find and run the default version according to your environment, providing flexibility.
>
>The downside is that in a multi-user system, someone else puts a bash in your $PATH, and an error may occur.
>
>In most cases, /usr/bin/env is the preferred choice, because it provides flexibility, especially if you want to run this script in a different version; and the way to specify the specific location#!/usr/bin/bash , In some cases it is safer because it limits the possibility of code injection.
>
>Maybe /usr/bin/bash does not exist on some systems, but /bin/bash must exist. So /bin/bash is the preferred choice for display designation.

dtmawo: I've just done copy-paste form other build scripts.
You're right, on the other hand:
- it's a build script intended to be executed in a specific environment
- this PR is not intended to fix all nuances of our toolchain

mswiatkowska (approved):

dtmawo:
> please remove binaries from git with rewriting history to not increasing repository size but `build/get_ext_tools.sh` run at build stage as name of this catalog tells

I'll give travis cache a try, if I can keep:
- builds reasonably fast (caching)
- security (sha256 validation)
then I'll try to remove binaries from git repo as suggested.

karolkieglerski (approved):

PR #277, opened 2022-01-28
[DO NOT MERGE] APM-340237: GCP Function - remove interactive params

mmajcher:
IMPORTANT

1. Change PR target to master after Marcin's embedded_ext_tools is merged
2. This needs to be synced with docs changes!!


scripts/setup.sh
```diff
@@ -212,54 +208,20 @@ SERVING_APP_ENGINE=$(echo "$APP_ENGINE" | "$JQ" -r '.servingStatus')
...
-      ;;
-    [Nn]*)
-      echo
-      echo -e "\e[91mERROR: \e[37mCannot continue without App Engine. Deployment aborted."
-      exit
-      ;;
-    *) echo "- please answer y or n" ;;
-    esac
-  done
+  echo "AppEngine not found. It will be created in your preferredAppEngineRegion $PREFERRED_APP_ENGINE_REGION"
```
karolkieglerski, line 211: no can't be like this. You can create only one App Engine on whole subscription and only one. If you will to in that way you can break customer subscription

mmajcher: But this line above "if [[ -z "$APP_ENGINE" ]]; then", this is a check for this, right? 
If there is app engine no action. If no app engine, then creating app engine. 
Am I missing something?

karolkieglerski: thats true but, APPEngine can be only one on whole subscription an after creation it can't be deleted. Also it could be in the same region but it is not required. better will be an error if not exits then creating by default in some region without client decision

mswiatkowska: + there is inconsistency in naming for two regions:
`Two locations, which are called europe-west and us-central in App Engine commands and in the Google Cloud Console, are called europe-west1 and us-central1`
https://cloud.google.com/appengine/docs/locations

That's why we list all available app engine regions with proper naming in console for client.

mmajcher: I talked to @karolkieglerski and we agreed on following solution:
1. add explicit functionRegion. Right now function implicitly takes region from cloud shell
2. AppEngine region - parameter stays, will add comment that this region is funky and will give link to AppEngine regions
3. The code for Appengine stays as it is now, i.e. if AppEngine not found, create one in preferredAppEngine region

We also talked about gcloud quiet by default (right now gcloud is still interactive; for example in our pipeline we use -d option of our script to make gcloud quiet, but default is to have gcloud interactive), but not sure if I will pursue it here.

@mswiatkowska does it look ok for you?

mswiatkowska: Yes, it looks ok to me. We considered using gcloud quiet by default but in the end we decided to leave interactive mode as default not to override e.g. roles quietly... not sure if it should be changed.

mmajcher: The role in 99.9% will be the default "dynatrace_extension.metrics", I think we can override it freely

karolkieglerski:
@mmajcher you must remember that if you changing this you also should change e2e tests

PR #280, opened 2022-02-01
[APM-345077] Memory leak in the GCP integration (monitoring function)

dtwwolinski:
- updated deployment of the function runtime to python 3.8
- updated environment variables to match missing ones


src/lib/context.py
```diff
@@ -139,7 +139,7 @@ def __init__(
         self.project_id_owner = project_id_owner
         self.dynatrace_api_key = dynatrace_api_key
         self.dynatrace_url = dynatrace_url
-        self.function_name = os.environ.get("FUNCTION_NAME", "Local")
+        self.function_name = os.environ.get("K_SERVICE", "Local")
```
karolkieglerski, line 142: why you are changing this?

scripts/setup.sh
```diff
@@ -182,6 +182,9 @@ GCP_ACCOUNT=$(gcloud config get-value account)
 echo -e "You are now logged in as [$GCP_ACCOUNT]"
 echo
 DEFAULT_PROJECT=$(gcloud config get-value project)
+GCP_REGION=$(gcloud config get-value functions/region)
```
karolkieglerski, line 185: @mmajcher you are adding this in other PR as parameter right?

src/lib/context.py
dtwwolinski: I've changed it because in the newer runtime the ````FUNCTION_NAME```` variable is not available. https://cloud.google.com/functions/docs/configuring/env-var#newer_runtimes

dtwwolinski: This is also a reason why I'd to add manually the 
````
GCP_PROJECT: '$GCP_PROJECT'
FUNCTION_REGION: '$GCP_REGION'
````
to environment variables list.

scripts/setup.sh
mmajcher: Yeah as a yaml parameter.
As I understand Wlad only needs to read this value, as such it will be easy for me to replace this code with my variable. So can stay as is and I will update this later in my PR

mmajcher (approved):

PR #281, opened 2022-02-03
APM-340237: GCP Function - remove interactive params

mmajcher:
Needs to be synced with doc change


mmajcher:
Sorry guys, I was on wrong branch earlier. Your comments regarding AppEngine region (some naming differences vs. normal GCP regions) are covered, 
both function_region and app_engine_region are in params, with short comment about naming differences

karolkieglerski (approved):

PR #284, opened 2022-02-09
APM-331487 check if monitoring.googleapis.com is enabled

karolkieglerski:


src/lib/gcp_apis.py
```diff
@@ -36,6 +36,10 @@ async def get_all_disabled_apis(context: MetricsContext, token: str, project_id:
                 disabled_services_json = await response.json()
                 disabled_services = disabled_services_json.get("services", [])
                 disabled_apis.update({disable_service.get("config", {}).get("name", "") for disable_service in disabled_services})
+                context.log(f"All disabled APIs {disabled_apis}")
```
mswiatkowska, line 39: I'm not sure if we want to log all disabled apis, if I remember correctly in the response there are many more apis that we don't support

src/main.py
```diff
@@ -135,6 +135,9 @@ async def handle_event(event: Dict, event_context, projects_ids: Optional[List[s
         disabled_apis = {}
         for project_id in projects_ids:
             disabled_apis = {project_id: await get_all_disabled_apis(context, token, project_id)}
+            if disabled_apis[project_id] == {project_id}:
+                projects_ids.remove(project_id)
+        context.log("monitoring.googleapis.com API enabled at following projects: " + ", ".join(projects_ids))
```
mswiatkowska, line 140: I'd prefer to have this log for projects where this api is disabled - logging only when something is wrong

src/main.py
```diff
@@ -135,6 +135,9 @@ async def handle_event(event: Dict, event_context, projects_ids: Optional[List[s
         disabled_apis = {}
         for project_id in projects_ids:
             disabled_apis = {project_id: await get_all_disabled_apis(context, token, project_id)}
+            if disabled_apis[project_id] == {project_id}:
+                projects_ids.remove(project_id)
+                context.log(f"monitoring.googleapis.com API disabled in the project: {project_id}, this project will not be monitored")
```
mswiatkowska, line 140: (minor) hope the customer won't have hundreds of such projects (with monitoring api disabled) ;) we could collect these project ids first and then log them once.

mswiatkowska (dismissed):

src/main.py
```diff
@@ -135,6 +135,9 @@ async def handle_event(event: Dict, event_context, projects_ids: Optional[List[s
         disabled_apis = {}
         for project_id in projects_ids:
             disabled_apis = {project_id: await get_all_disabled_apis(context, token, project_id)}
+            if disabled_apis[project_id] == {project_id}:
+                projects_ids.remove(project_id)
```
mswiatkowska, line 139: I'm afraid we can't remove project_id during iteration - I'm not sure what Python does in this case... as I mentioned in the comment about logging, I would collect all projects with disabled monitoring api and after that remove them from projects_ids and log

mswiatkowska (dismissed):

src/main.py
```diff
@@ -133,8 +133,16 @@ async def handle_event(event: Dict, event_context, projects_ids: Optional[List[s
             projects_ids = await get_all_accessible_projects(context, gcp_session, token)
 
         disabled_apis = {}
+        disabled_projects = []
         for project_id in projects_ids:
             disabled_apis = {project_id: await get_all_disabled_apis(context, token, project_id)}
+            if disabled_apis[project_id] == {project_id}:
```
Dawid-Kaszubski, line 139: Can we have check if `monitoring.googleapis.com` is in disabled_apis here? 
IMHO additional logic with passing only project_id in set is unnecessary.

Dawid-Kaszubski (changes requested):

Dawid-Kaszubski (dismissed):

mswiatkowska (approved):

Dawid-Kaszubski (approved):

PR #287, opened 2022-02-16
APM-326214 cloud.log_forwarder attribute was added to distinguish different setups.

IrynaKudlaienko:


karolkieglerski (approved):

tests/unit/extraction_rules/test_cloud_log_forwarder.py
```diff
@@ -0,0 +1,84 @@
...
+    # given
+    monkeypatch.setattr(logs_processor, 'CLOUD_LOG_FORWARDER', "")
+    monkeypatch.setattr(logs_processor, 'CLOUD_LOG_FORWARDER_POD', "myPod")
+
+    # when
+    actual_output = _create_dt_log_payload(TEST_LOGS_PROCESSING_CONTEXT, json.dumps(test_record))
+
+    # then
+    if not actual_output.get('cloud.log_forwarder'):
+        assert True
```
wojtasskorcz, line 71: This way this test will never fail (assert will be skipped in a failure scenario). We should assert on something like
```
assert actual_output.get('cloud.log_forwarder').isempty()
```
or something similar. Same with the test below

wojtasskorcz (approved):

karolkieglerski (approved):

PR #292, opened 2022-03-25
APM-357141 Enable GKE API if it is not enabled

joaquinfilipic-dynatrace:
During installation of autopilot cluster, check if the GKE API needs to be enabled

### DIFF
--- a/scripts/deploy-helm.sh
+++ b/scripts/deploy-helm.sh
@@ -359,9 +359,9 @@ if [[ $CREATE_AUTOPILOT_CLUSTER == "Y" ]]; then
     err "Default region not set. Set default region by running 'gcloud config set compute/region <REGION>'."
     exit 1
   fi
-  CONTAINERS_API=$(gcloud services list --enabled --filter=container.googleapis.com --project="$GCP_PROJECT")
-  if [[ -z "$CONTAINERS_API" ]]; then
-    debug "Enabling Containers API"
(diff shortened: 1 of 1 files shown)


scripts/deploy-helm.sh
```diff
@@ -359,6 +359,11 @@ if [[ $CREATE_AUTOPILOT_CLUSTER == "Y" ]]; then
     err "Default region not set. Set default region by running 'gcloud config set compute/region <REGION>'."
     exit 1
   fi
+  CONTAINERS_API=$(gcloud services list --enabled --filter=container.googleapis.com --project="$GCP_PROJECT")
+  if [[ -z "$CONTAINERS_API" ]]; then
+    debug "Enabling Containers API"
```
equardo, line 364: Shouldn't this be Kubernetes Engine API?

joaquinfilipic-dynatrace: Yes, that's the name Google uses for the service:
https://cloud.google.com/kubernetes-engine/docs/reference/rest/#:~:text=Kubernetes%20Engine%20API%20Builds%20and%20manages%20container-based,applications%2C%20powered%20by%20the%20open%20source%20Kubernetes%20technology.

equardo: I mean that the debug message says "Enabling Containers API"

mswiatkowska (approved):

scripts/deploy-helm.sh
joaquinfilipic-dynatrace: Ok, changed to be more accurate

equardo (approved):

PR #294, opened 2022-03-31
Increase interval and waiting time to 7 min

joaquinfilipic-dynatrace:


tests/e2e/execute-tests-cloud-function.sh
```diff
@@ -17,7 +17,7 @@ START_LOAD_GENERATION=$(date -u +%s%3N)
 export START_LOAD_GENERATION
 
 ./tests/e2e/deployment-test-cloud-function.sh
-sleep 420
+sleep 600
```
kamiljsokolowski, line 20: Those `sleep` statements are fine but, in the future, we should be aiming for some kind of verification/check procedures if script execution results are in effect (i.e. call GCP API if cloud function is deployed).

kamiljsokolowski (approved):

tests/e2e/execute-tests-cloud-function.sh
joaquinfilipic-dynatrace: Yes, right. For now, I was just trying different values, but I need to push them to test. If any change could be made, I will discuss it first.

wojtasskorcz:
Still didn't help with the branch building or is it because we still don't have Travis license/credits?

PR #295, opened 2022-04-04
Bugfix/apm 363386 gke deployment respect service account param

mmajcher:
Script used to have --service-acount param which was not really respected. SA with given name was created but helm deployment used hardcoded default value everywhere. 
Now param is extracted to values.yaml and is respected both in SA creation during deployment and SA assingment in helm templates.

### DIFF
--- a/k8s/helm-chart/dynatrace-gcp-function/templates/function-deployment.yml
+++ b/k8s/helm-chart/dynatrace-gcp-function/templates/function-deployment.yml
@@ -15,10 +15,10 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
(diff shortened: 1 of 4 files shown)


equardo (commented):
So I understand that now service account name is pulled from values.yaml. Will you change https://www.dynatrace.com/support/help/shortlink/existing-gke#param and https://www.dynatrace.com/support/help/shortlink/existing-gke#script?

dtwwolinski (approved):

mmajcher:
Yes, I will also change the docs

Dawid-Kaszubski (approved):

PR #296, opened 2022-04-05
Bugfix/apm 362265 gcp get disabled apis call loops forever

equardo:
Pagination in get disabled APIs call continued indefinetely on non-200 responses.
Additionally, each loop, an additional pageToken param was appended, creating impossibly long URLs, leading to 400s and 413s.
This led to the endpoint being always busy with too many calls.

### DIFF
--- a/src/lib/gcp_apis.py
+++ b/src/lib/gcp_apis.py
@@ -24,14 +24,12 @@ async def get_all_disabled_apis(context: MetricsContext, token: str, project_id:
     disabled_apis = set()
     try:
         response = await context.gcp_session.get(base_url, headers=headers, raise_for_status=True)
-        context.log(f'Initial response {response}')
         disabled_services_json = await response.json()
         disabled_services = disabled_services_json.get("services", [])
         disabled_apis.update({disable_service.get("config", {}).get("name", "") for disable_service in disabled_services})
         while disabled_services_json.get("nextPageToken"):
             url = f"{base_url}&pageToken={disabled_services_json['nextPageToken']}"
             response = await context.gcp_session.get(url, headers=headers, raise_for_status=True)
-            context.log(f'Continued response {response}')
             disabled_services_json = await response.json()
             disabled_services = disabled_services_json.get("services", [])
             disabled_apis.update({disable_service.get("config", {}).get("name", "") for disable_service in disabled_services})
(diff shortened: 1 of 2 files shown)


joaquinfilipic-dynatrace (approved):

mswiatkowska (approved):

src/lib/gcp_apis.py
```diff
@@ -11,32 +11,32 @@
...
     try:
-        response = await context.gcp_session.get(url, headers=headers)
-        if response.status != 200:
-            context.log(f'Http error: {response.status}, url: {response.url}, reason: {response.reason}')
-            return set()
+        response = await context.gcp_session.get(base_url, headers=headers, raise_for_status=True)
         disabled_services_json = await response.json()
         disabled_services = disabled_services_json.get("services", [])
         disabled_apis.update({disable_service.get("config", {}).get("name", "") for disable_service in disabled_services})
         while disabled_services_json.get("nextPageToken"):
```
wojtasskorcz, line 30: I don't understand it. You say that for non-200 responses the code looped forever. Was it on this while-loop? Because if there's non-200 response we shouldn't even have reached this loop (we would've returned empty set in line 28).

equardo: That is true, and if the first call fails, we don't get disabled APIs, continue and everything is fine.
However, oftentimes the initial call succeeds, yet most of the following ones do not leading to the URL becoming too long and eventually no requests passing at all and the loop becoming infinite. 
One approach to consider could be only changing the URL param appending, but I'm not sure if there is much upside - if we sometimes fail because of a temporary issue, we'll just do some more requests than needed later to the disabled APIs.

equardo: That is true, and if the first call fails, we don't get disabled APIs, continue and everything is fine.
However, oftentimes the initial call succeeds, yet most of the following ones do not leading to the URL becoming too long and eventually no requests passing at all and the loop becoming infinite.
One approach to consider could be only changing the URL param appending, but I'm not sure if there is much upside - if we sometimes fail because of a temporary issue, we'll just do some more requests than needed later to the disabled APIs.

wojtasskorcz: Ahh, I see, because `disabled_services_json` is only updated when we get 200 in the loop. Otherwise we reuse the value that we received in the initial request all the time

wojtasskorcz (approved):

Dawid-Kaszubski (approved):

PR #309, opened 2022-07-04
Apm 378148 log access token public part

joaquinfilipic-dynatrace:


tests/unit/test_fast_check.py
```diff
@@ -24,7 +24,7 @@ def test_token_masking_old_format():
 def test_token_masking_new_format():
     assert_that(obfuscate_dynatrace_access_key(
         "dt0c01.FAKE11.FAKEEU7QQ5KW7BZVN2XQFAKEFAKEJFHLZKTOUTFZKAAZTRSFR4HTJXZKXIZPPUVZ2D7YPRXY2IUFAKE7JY")).is_equal_to(
-        "dt0c01.FAK***********************************************************************************7JY")
+        "FAKE11")
```
dt-be, line 27: I'd expect to log 'dt0c01.FAKE11' - that's what we do elsewhere (cluster UI, debugUI, restAPI) & it does make you easily recognize this is the token + that it us the new format token.

joaquinfilipic-dynatrace: Ok, will do it

src/lib/fast_check.py
```diff
@@ -119,7 +119,8 @@ def obfuscate_dynatrace_access_key(dynatrace_access_key: str):
     if len(dynatrace_access_key) >= 7:
         is_new_token = dynatrace_access_key.startswith("dt0c01.") and len(dynatrace_access_key) == 96
         if is_new_token:
-            return dynatrace_access_key[:10] + '*' * (len(dynatrace_access_key) - 13) + dynatrace_access_key[-3:]
+            # characters between dots are the public part of the token
+            return dynatrace_access_key.split('.')[1]
```
equardo, line 123: So here we skip the "dtc01." part, correct? Wouldn't it be better to keep it just to make it clearer what this is?

equardo: Sorry about the after-merge comment but you speedran the merge 😅

joaquinfilipic-dynatrace: Haha my fault, I will wait till the next day next time

PR #310, opened 2022-07-05
Fix: include token prefix

joaquinfilipic-dynatrace:


src/lib/fast_check.py
```diff
@@ -120,7 +120,7 @@ def obfuscate_dynatrace_access_key(dynatrace_access_key: str):
         is_new_token = dynatrace_access_key.startswith("dt0c01.") and len(dynatrace_access_key) == 96
         if is_new_token:
             # characters between dots are the public part of the token
-            return dynatrace_access_key.split('.')[1]
+            return 'dt0c01.' + dynatrace_access_key.split('.')[1]
```
dt-be, line 123: Oh, common, log what's actually there in [0] :D

joaquinfilipic-dynatrace: But that's faster, haha. I wrote that because of the previous condition that's already checking that hardcoded prefix

dtwwolinski (approved):

PR #313, opened 2022-07-12
Fix conditions of parameters validations and condition of invalid token permissions

joaquinfilipic-dynatrace:
Changes:
- Set USE_EXISTING_ACTIVE_GATE to true by default before readonly statement.
- DYNATRACE_URL validations in every case, since it's now required.
- LOG_VIEWER URL for every case in which logs are ingested, since DYNATRACE_URL is now required.
- Log record indicating invalid token permissions is now displayed only in case the token was retrieved successfully but has incorrect scopes (or is revoked). Before this, if a network/communication error happened in the token call, the token_metadata variable was returned as an empty dict ({}) and this log was also displayed, which was misleading.
- Remove occurrences of whitespaces, \r, \n & \t in error prone parameters

### DIFF
--- a/scripts/deploy-helm.sh
+++ b/scripts/deploy-helm.sh
@@ -163,11 +163,11 @@ done
 
 GCP_PROJECT=$(helm show values ./dynatrace-gcp-function --jsonpath "{.gcpProjectId}")
 DEPLOYMENT_TYPE=$(helm show values ./dynatrace-gcp-function --jsonpath "{.deploymentType}")
-DYNATRACE_ACCESS_KEY=$(helm show values ./dynatrace-gcp-function --jsonpath "{.dynatraceAccessKey}")
+DYNATRACE_ACCESS_KEY=$(helm show values ./dynatrace-gcp-function --jsonpath "{.dynatraceAccessKey}" | sed 's/[\r\n\t ]//g')
 readonly DYNATRACE_ACCESS_KEY
-DYNATRACE_URL=$(helm show values ./dynatrace-gcp-function --jsonpath "{.dynatraceUrl}" | sed 's:/*$::')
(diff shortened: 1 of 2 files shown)


equardo (dismissed):

Dawid-Kaszubski (dismissed):

scripts/deploy-helm.sh
```diff
@@ -163,16 +163,19 @@ done
...
 GCP_PROJECT=$(helm show values ./dynatrace-gcp-function --jsonpath "{.gcpProjectId}")
 DEPLOYMENT_TYPE=$(helm show values ./dynatrace-gcp-function --jsonpath "{.deploymentType}")
-DYNATRACE_ACCESS_KEY=$(helm show values ./dynatrace-gcp-function --jsonpath "{.dynatraceAccessKey}")
+DYNATRACE_ACCESS_KEY=$(helm show values ./dynatrace-gcp-function --jsonpath "{.dynatraceAccessKey}" | sed 's/[\r\n\t ]//g')
 readonly DYNATRACE_ACCESS_KEY
-DYNATRACE_URL=$(helm show values ./dynatrace-gcp-function --jsonpath "{.dynatraceUrl}" | sed 's:/*$::')
+DYNATRACE_URL=$(helm show values ./dynatrace-gcp-function --jsonpath "{.dynatraceUrl}" | sed 's/[\r\n\t ]//g' | sed 's:/*$::')
 readonly DYNATRACE_URL
-DYNATRACE_LOG_INGEST_URL=$(helm show values ./dynatrace-gcp-function --jsonpath "{.dynatraceLogIngestUrl}" | sed 's:/*$::')
+DYNATRACE_LOG_INGEST_URL=$(helm show values ./dynatrace-gcp-function --jsonpath "{.dynatraceLogIngestUrl}" | sed 's/[\r\n\t ]//g' | sed 's:/*$::')
```
equardo, line 170: Not very familiar with sed, can you explain why use two separate sed commands?

Dawid-Kaszubski (approved):

equardo (approved):

PR #315, opened 2022-07-25
Added vpcNetwork parameter

joaquinfilipic-dynatrace:
Added vpcNetwork param to values.yaml, with default value, so the customer can select in which network to deploy the autopilot cluster

### DIFF
--- a/k8s/helm-chart/dynatrace-gcp-function/values.yaml
+++ b/k8s/helm-chart/dynatrace-gcp-function/values.yaml
@@ -57,7 +57,7 @@ activeGate:
 
 # COMMON VALUES - DEFAULTS
 
-# vcpNetwork: name of an existing VPC Network in which the autopilot cluster will be deployed. By default, the network named 'default' is used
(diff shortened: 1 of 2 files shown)


equardo (approved):

k8s/helm-chart/dynatrace-gcp-function/values.yaml
```diff
@@ -57,6 +57,8 @@ activeGate:
 
 # COMMON VALUES - DEFAULTS
 
+# vcpNetwork: name of an existing VPC Network in which the autopilot cluster will be deployed. By default, the network named 'default' is used
```
tepesware, line 60:
```suggestion
# vcpNetwork: name of an existing VPC Network in which the autopilot cluster will be deployed. By default or when empty, the network named 'default' is used
```

tepesware (changes requested):

tepesware (approved):
It's me Wlad ;)

Dawid-Kaszubski (approved):

equardo (approved):

PR #335, opened 2022-11-30
PCLOUDS-1523 utilized asyncio to iterate through projects to find val…

NematulloKozimov:
…id ones

### DIFF
--- a/scripts/deploy-helm.sh
+++ b/scripts/deploy-helm.sh
@@ -367,6 +367,9 @@ if [[ $CREATE_AUTOPILOT_CLUSTER == "Y" ]]; then
   info ""
   info "- Create and connect GKE Autopilot k8s cluster ${AUTOPILOT_CLUSTER_NAME}."
   gcloud components install gke-gcloud-auth-plugin
+  export USE_GKE_GCLOUD_AUTH_PLUGIN=True
+  source ~/.bashrc
+  gcloud components update
   gcloud container clusters create-auto "${AUTOPILOT_CLUSTER_NAME}" --project "${GCP_PROJECT}" --zone "" --network "${VPC_NETWORK}" | tee -a "$FULL_LOG_FILE" >${CMD_OUT_PIPE}
   gcloud container clusters get-credentials "${AUTOPILOT_CLUSTER_NAME}" --project "${GCP_PROJECT}" --zone "" | tee -a "$FULL_LOG_FILE" >${CMD_OUT_PIPE}
 fi
--- a/scripts/deploy-helm.sh
+++ b/scripts/deploy-helm.sh
@@ -366,6 +366,7 @@ if [[ $CREATE_AUTOPILOT_CLUSTER == "Y" ]]; then
   fi
   info ""
   info "- Create and connect GKE Autopilot k8s cluster ${AUTOPILOT_CLUSTER_NAME}."
+  export USE_GKE_GCLOUD_AUTH_PLUGIN=True
   gcloud container clusters create-auto "${AUTOPILOT_CLUSTER_NAME}" --project "${GCP_PROJECT}" --zone "" --network "${VPC_NETWORK}" | tee -a "$FULL_LOG_FILE" >${CMD_OUT_PIPE}
   gcloud container clusters get-credentials "${AUTOPILOT_CLUSTER_NAME}" --project "${GCP_PROJECT}" --zone "" | tee -a "$FULL_LOG_FILE" >${CMD_OUT_PIPE}
 fi
--- a/scripts/deploy-helm.sh
+++ b/scripts/deploy-helm.sh
@@ -368,7 +368,6 @@ if [[ $CREATE_AUTOPILOT_CLUSTER == "Y" ]]; then
   info "- Create and connect GKE Autopilot k8s cluster ${AUTOPILOT_CLUSTER_NAME}."
   gcloud components install gke-gcloud-auth-plugin
   export USE_GKE_GCLOUD_AUTH_PLUGIN=True
-  source ~/.bashrc
   gcloud components update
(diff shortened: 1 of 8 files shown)


src/lib/fast_check.py
```diff
@@ -191,20 +208,14 @@ async def _check_services(self, project_id):
...
 
-            results = await asyncio.gather(*tasks, return_exceptions=True)
-
-            if all(result is not None for result in results):
-                ready_to_monitor.append(project_id)
+        results = await asyncio.gather(*tasks, return_exceptions=True)
+        if all(result is not None for result in results):
+            ready_to_monitor.append(project_id)
 
+        if K8S_CONTAINER_NAME_PREFIX not in HOSTNAME:
```
mmajcher, line 218: If this is some way to check "am i in container or in function" then please make this a nice-named function and put it in some good place

src/lib/fast_check.py
```diff
@@ -127,15 +138,18 @@ def obfuscate_dynatrace_access_key(dynatrace_access_key: str):
...
 
 
-async def check_dynatrace(logging_context: LoggingContext, project_id, dt_session: ClientSession, dynatrace_url, dynatrace_access_key):
+async def check_dynatrace(logging_context: LoggingContext, project_id, dt_session: ClientSession, dynatrace_url,
+                          dynatrace_access_key):
     try:
         if not dynatrace_url or not dynatrace_access_key:
-            logging_context.log(f'ERROR No Dynatrace secrets: DYNATRACE_URL, DYNATRACE_ACCESS_KEY for project: {project_id}.'
-                                     f'Add required secrets to Secret Manager.')
+            logging_context.log(
```
mmajcher, line 145: many changes to formatting, are they really needed?

src/lib/fast_check.py
```diff
@@ -16,6 +16,9 @@
 import json
 import os
 import re
+import time
+import threading
+from concurrent.futures import ThreadPoolExecutor
```
mmajcher, line 21: why??

joaquinfilipic-dynatrace (approved):

src/lib/fast_check.py
```diff
@@ -191,20 +192,14 @@ async def _check_services(self, project_id):
...
-
-            results = await asyncio.gather(*tasks, return_exceptions=True)
-
-            if all(result is not None for result in results):
-                ready_to_monitor.append(project_id)
+        results = await asyncio.gather(*tasks, return_exceptions=True)
+        if all(result is not None for result in results):
+            ready_to_monitor.append(project_id)
 
+        if not is_deployment_running_inside_gke_container():
```
mmajcher, line 202: I would go more with "is_deployment_running_inside_cloud_function", it is kinda same but this version is more explicit. It will be easier to spot it and remove when we get to cutting out cloud function option

src/lib/fast_check.py
```diff
@@ -191,20 +192,14 @@ async def _check_services(self, project_id):
...
-
-        ready_to_monitor = []
-        for project_id in project_list:
-            tasks = [self._check_services(project_id)]
-
-            results = await asyncio.gather(*tasks, return_exceptions=True)
-
-            if all(result is not None for result in results):
-                ready_to_monitor.append(project_id)
+        results = await asyncio.gather(*tasks, return_exceptions=True)
```
mmajcher, line 198: I think it is not obvious what return one should expect from this list of tasks that you get from check_services.
Not sure how to express it well

Bc I am thinking about this "if all" here.
Do we need result!=None for all of those tasks? If 10 is ok and 1 is None then we don't monitor project? Can you explain this, this probably needs a comment of explanation in the code

src/lib/fast_check.py
```diff
@@ -213,6 +208,18 @@ async def execute(self) -> FastCheckResult:
...
 
+    async def execute(self) -> FastCheckResult:
+        _check_configuration_flags(self.logging_context, METRICS_CONFIGURATION_FLAGS)
+
+        project_list = await get_all_accessible_projects(self.logging_context, self.gcp_session, self.token)
+
+        ready_to_monitor = []
+        tasks_to_find_ready_projects = []
+        for project_id in project_list:
+            tasks_to_find_ready_projects.append(self.is_project_ready_to_monitor(project_id, ready_to_monitor))
```
mmajcher, line 219: is async routine adding stuff to an array that is here (level above)? This is a little messy design, is this much work to have async routine return result like (<PROJECT_ID>, <CAN_MONITOR=TRUE/FALSE>) and only take IDs from the results that we get from asyncio gather?
Side effects (modifying external objects) should be limited to minimum where possible

src/lib/fast_check.py
mmajcher: Like - what the hell is check services and why we need all results to be true?

joaquinfilipic-dynatrace:
Perhaps you've already discussed this, but if we are doing extra refactors (not only fixing the bug or adding comments) we need to spend more time testing it carefully and/or contemplating changes.

joaquinfilipic-dynatrace (dismissed):

scripts/deploy-helm.sh
```diff
@@ -366,6 +366,7 @@ if [[ $CREATE_AUTOPILOT_CLUSTER == "Y" ]]; then
   fi
   info ""
   info "- Create and connect GKE Autopilot k8s cluster ${AUTOPILOT_CLUSTER_NAME}."
+  export USE_GKE_GCLOUD_AUTH_PLUGIN=True
```
mmajcher, line 369: Explanation for this?

mmajcher (dismissed):

mmajcher (approved):

joaquinfilipic-dynatrace (approved):

PR #343, opened 2022-12-15
PCLOUDS-1532 Used concurrent queries for initial project checkups

NematulloKozimov:


src/main.py
```diff
@@ -168,6 +173,15 @@ async def handle_event(event: Dict, event_context, projects_ids: Optional[List[s
     # Noise on windows at the end of the logs is caused by https://github.com/aio-libs/aiohttp/issues/4324
 
 
+async def check_if_project_is_disabled_and_get_disabled_api_set(context: MetricsContext, project_id: str):
+    await check_x_goog_user_project_header_permissions(context, project_id)
+    disabled_api_set = {project_id: await get_all_disabled_apis(context, project_id)}
```
joaquinfilipic-dynatrace, line 178: Maybe I'm missing something, but why is it here a new set each time instead of updating directly disabled_apis? And later we're iterating through results of tasks again to do that?

src/main.py
```diff
@@ -168,6 +174,15 @@ async def handle_event(event: Dict, event_context, projects_ids: Optional[List[s
     # Noise on windows at the end of the logs is caused by https://github.com/aio-libs/aiohttp/issues/4324
 
 
+async def check_if_project_is_disabled_and_get_disabled_api_set(context: MetricsContext, project_id: str) -> [str, bool, set]:
```
joaquinfilipic-dynatrace, line 177: I believe we are returning "Tuple[str, bool, Set[...]]", aren't we?
https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html

mmajcher: I think it's the same, in python using a comma is basically defining a tuple?

joaquinfilipic-dynatrace: Yes. And [] would be a list

joaquinfilipic-dynatrace: It changes nothing, but type hints are simply that, hints

mmajcher (approved):

PR #344, opened 2022-12-16
Mw fixes

dtmawo:


src/lib/gcp_apis.py
```diff
@@ -11,15 +11,47 @@
...
+            if next_token:
+                query_params["pageToken"] = next_token
+            response = await context.gcp_session.get(
+                GCP_SERVICE_USAGE_URL + f'/projects/{project_id}/services',
+                headers={
+                    "Authorization": f'Bearer {context.token}'
+                },
+                params=query_params)
+            if response.status != 200:
+                context.logging_context.log(
```
mmajcher, line 39: I think logging_context.error is there for this

src/main.py
```diff
@@ -76,16 +69,8 @@ def is_yaml_file(f: str) -> bool:
...
-    if isinstance(event_context, Dict):
-        # for k8s installation
-        context = LoggingContext(event_context.get("execution_id", None))
-    else:
-        context = LoggingContext(None)
-
-    if not services:
-        # load services for GCP Function
-        services = load_supported_services(context)
+async def query_metrics(execution_id: Optional[str], services: Optional[List[GCPService]] = None):
```
mmajcher, line 72: I think handle_event is execution start point for cloud function? And probably we need such function for now

PR #346, opened 2022-12-20
Get projects in every monitoring loop + Refactor

joaquinfilipic-dynatrace:
- Get projects in every monitoring loop. Deleted project_ids parameter.
- Remove code for asking for projects and services during fastcheck. It's done directly on every monitoring loop.
- pageSize = 200 instead of default 50 in request for services.
- Pod resources values based on the running integrations in dynatrace-gcp-extension. For autopilot clusters, only requests are considered, so limits were removed: https://cloud.google.com/kubernetes-engine/docs/concepts/autopilot-resource-requests#resource-limits
- Increased timeout for livenessProbe and startupProbe (default was 1s). We can still adjust these values.
- Minor refactors, mainly based on Marcin's PR.

There are still things to improve/fix, but in separate PRs.


src/main.py
```diff
@@ -40,34 +40,28 @@
...
     except Exception as e:
         traceback.print_exc()
         raise e
 
 
-async def async_dynatrace_gcp_extension(project_ids: Optional[List[str]] = None, services: Optional[List[GCPService]] = None):
+async def async_dynatrace_gcp_extension(services: Optional[List[GCPService]] = None):
     """
-    Used in docker or for tests
+    Starting point for installation as an autopilot cluster and for tests.
```
mmajcher, line 55: 1 I think this code is not limited to autopilot, you can have your existing GKE
2 This code is not really k8s specific, let's focus on the fact that it is for "long-running" or "container"

src/main.py
```diff
@@ -76,15 +70,10 @@ def is_yaml_file(f: str) -> bool:
...
-        # for k8s installation
-        context = LoggingContext(event_context.get("execution_id", None))
-    else:
-        context = LoggingContext(None)
-
+async def query_metrics(execution_id: Optional[str], services: Optional[List[GCPService]] = None):
+    context = LoggingContext(execution_id)
     if not services:
-        # load services for GCP Function
+        # Load services for GCP Function and for tests
```
mmajcher, line 76: why do we add this "and for tests", what does it mean??

src/main.py
```diff
@@ -129,13 +118,13 @@ async def handle_event(event: Dict, event_context, projects_ids: Optional[List[s
...
-        if not projects_ids:
-            projects_ids = await get_all_accessible_projects(context, gcp_session, token)
+        projects_ids = await get_all_accessible_projects(context, gcp_session, token)
 
         disabled_projects, disabled_apis_by_project_id = await get_disabled_projects_and_disabled_apis_by_project_id(context, projects_ids)
 
         if disabled_projects:
-            context.log(f"monitoring.googleapis.com API disabled in the projects: " + ", ".join(disabled_projects) + ", that projects will not be monitored")
+            context.log(f"Cannot monitor projects: {disabled_projects}. "
+                        f"Enable required services to do so: {REQUIRED_SERVICES}")
```
mmajcher, line 127: Is this the only reason (not having required services) for projects being disabled?

mmajcher: Even if it is, I would log it somewhere near that logic. Because if this logic changes, I dont think anyone would remember about this log line here that gives reason

src/main.py
```diff
@@ -159,7 +148,7 @@ async def handle_event(event: Dict, event_context, projects_ids: Optional[List[s
         await gcp_session.close()
         await dt_session.close()
 
-    # Noise on windows at the end of the logs is caused by https://github.com/aio-libs/aiohttp/issues/4324
+    # Noise on Windows at the end of the logs is caused by https://github.com/aio-libs/aiohttp/issues/4324
```
mmajcher, line 151: :D

src/main.py
joaquinfilipic-dynatrace: Now at least, yes

src/main.py
joaquinfilipic-dynatrace: Tests call the function without the services param. Nevertheless, this is one of the things to change, probably.

src/main.py
joaquinfilipic-dynatrace: Right. I'll change the comment later too

NematulloKozimov (approved):

mmajcher (approved):

dtwwolinski (commented):
+1

PR #351, opened 2022-12-30
PCLOUDS-1533: Keep task reference in logs SFM loop

joaquinfilipic-dynatrace:


src/lib/logs/log_self_monitoring.py
```diff
@@ -61,11 +61,15 @@ def put_sfm_into_queue(context: LogsContext):
 
 
 async def create_sfm_worker_loop(sfm_queue: Queue, logging_context: LoggingContext, instance_metadata: InstanceMetadata):
+    sfm_loop = asyncio.get_event_loop()
```
mmajcher, line 64: Is this really sfm_loop, like only for sfm? Or is it some main loop that you reuse here. If so, then the naming is off

src/lib/logs/log_self_monitoring.py
```diff
@@ -61,11 +61,15 @@ def put_sfm_into_queue(context: LogsContext):
...
+    sfm_loop_tasks = set()
     while True:
         try:
             await asyncio.sleep(SFM_WORKER_EXECUTION_PERIOD_SECONDS)
             self_monitoring = LogSelfMonitoring()
-            asyncio.get_event_loop().create_task(_loop_single_period(self_monitoring, sfm_queue, logging_context, instance_metadata))
+            sfm_loop_task = sfm_loop.create_task(_loop_single_period(self_monitoring, sfm_queue, logging_context, instance_metadata))
+            sfm_loop_tasks.add(sfm_loop_task)
```
mmajcher, line 71: Please add a short comment -> you need to keep some reference to the task otherwise it might get garbage collected, see docs

src/lib/logs/log_self_monitoring.py
joaquinfilipic-dynatrace: Right. Main loop

mmajcher (approved):

PR #354, opened 2023-01-03
PCLOUDS-1494: Metrics scope support

joaquinfilipic-dynatrace:
- Added param to enable metrics scope feature. Check disabled APIs and search topology are skipped in this case.
- Added test for e2e-helm-deployment-test stage, to check metrics from a service present in another project.
- Adapted e2e-cloud-function-deployment-test stage, because it uses the same test as the previous stage.
- Minor refactors


NematulloKozimov (approved):

scripts/deploy-helm.sh
```diff
@@ -421,7 +421,7 @@ info ""
 info "- 4. Create dynatrace-gcp-function IAM role(s)."
 if [[ $DEPLOYMENT_TYPE == logs ]] || [[ $DEPLOYMENT_TYPE == all ]]; then
   debug "Creating or updating GCP IAM role with gcp_iam_roles/dynatrace-gcp-function-logs-role.yaml"
-  if [[ $(gcloud iam roles list --filter="name:$ROLE_NAME.logs" --project="$GCP_PROJECT" --format="value(name)") ]]; then
+  if [[ $(gcloud iam roles list --filter="name ~ $ROLE_NAME.logs" --project="$GCP_PROJECT" --format="value(name)") ]]; then
```
joaquinfilipic-dynatrace, line 424: To avoid future warnings:
WARNING: --filter : operator evaluation is changing for consistency across Google APIs

Ref: https://cloud.google.com/sdk/gcloud/reference/topic/filters

src/main.py
```diff
@@ -149,10 +154,6 @@ async def query_metrics(execution_id: Optional[str], services: Optional[List[GCP
     # Noise on Windows at the end of the logs is caused by https://github.com/aio-libs/aiohttp/issues/4324
 
 
-def is_yaml_file(f: str) -> bool:
```
joaquinfilipic-dynatrace, line 152: Moved to utilities.py only

src/main.py
```diff
@@ -259,32 +265,25 @@ def load_supported_services(context: LoggingContext) -> List[GCPService]:
...
-        context.log("Selected feature sets: " + ", ".join(featureSets))
+    feature_sets = [f"{service.name}/{service.feature_set}" for service in services]
+    if feature_sets:
+        context.log("Selected feature sets: " + ", ".join(feature_sets))
     else:
         context.log("Empty feature sets. GCP services not monitored.")
     return services
 
 
-def extract_technology_name(config_yaml):
```
joaquinfilipic-dynatrace, line 281: Moved to utilities.py only

src/lib/fast_check.py
```diff
@@ -76,10 +77,6 @@
 FastCheckResult = NamedTuple('FastCheckResult', [('projects', List[str])])
 
 
-def find_service_name(service):
```
joaquinfilipic-dynatrace, line 79: Not used, didn't find any occurrence

activation-config.yaml
```diff
@@ -56,6 +56,10 @@ googleCloud:
     # Send custom metrics to GCP to diagnose quickly if your dynatrace-gcp-function processes and sends metrics to Dynatrace properly.
     # Allowed values: true/yes, false/no
     selfMonitoringEnabled: false
+    # scopingProjectSupportEnabled: set to true when metrics scope is configured. Metrics will be collected from all monitored projects,
+    # but topology fetch will not, so we need to bypass the verification in case there are instances of a service in monitored projects
```
mmajcher, line 60: Is this explanation of behavior changes needed here? I can see us giving more tweaks based on this variable and probably we will not remember to update this var description

activation-config.yaml
```diff
@@ -56,6 +56,10 @@ googleCloud:
...
+    # and not in the scoping project.
```
mmajcher, line 61: What happens if you have metric scope for other project but this variable is set to false by mistake?

src/lib/fast_check.py
```diff
@@ -160,7 +157,7 @@ def __init__(self, gcp_session: ClientSession, dt_session: ClientSession, token:
         self.logging_context = logging_context
         self.token = token
 
-    async def is_project_ready_to_monitor(self, project_id):
+    async def is_project_ready_to_monitor(self, project_id: str) -> Tuple[str, bool]:
```
mmajcher, line 160: I am not sure why we are returning parameter, I mean you get project_id and return project_id??? This signature seems weird.

src/lib/utilities.py
```diff
@@ -86,3 +97,8 @@ def print_dynatrace_logo():
     print("  ,     ,,,,,,,,,,,,,,,,,,,,,,,,,,,,,  ,,,,,,,")
     print("     ,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,,  ,,,,")
     print("")
+
+
+class ConfigurationParameters:
```
mmajcher, line 102: lib/configuration/config.py

src/main.py
```diff
@@ -172,22 +173,27 @@ async def fetch_ingest_lines_task(context: MetricsContext, project_id: str, serv
...
-        skipped_topology_services_string = ", ".join(skipped_topology_services)
-        context.log(project_id, f"Skipped fetching topology for disabled services: {skipped_topology_services_string}")
-
-    fetch_topology_results = await asyncio.gather(*topology_tasks, return_exceptions=True)
-
+    fetch_topology_results = ()
+
+    # Using metrics scope feature, fetching topology is not needed
+    if not ConfigurationParameters.scoping_project_support_enabled:
+        for service in services:
```
mmajcher, line 180: This block looks like candidate for method extraction

activation-config.yaml
joaquinfilipic-dynatrace: This is uncertain, because if you have instances of a service in the scoping project, you will collect metrics from the other monitored projects as well, because it will pass the topology search. I don't know if there's anything to be done in that case, to me it is a misconfiguration, because you set the GCP env/main project to collect metrics from other projects, but you put false here.

src/main.py
joaquinfilipic-dynatrace: There are some minor impl alternatives here, but I extracted it.

src/lib/fast_check.py
joaquinfilipic-dynatrace: This is likely due to how we are iterating the returning params later, we need the relation project_id - is_project_ready. And it's a class method (MetricsFastCheck), not a general one, but if you are strongly against it, I can try and refactor it as well.

NematulloKozimov (dismissed):

mmajcher (approved):

mmajcher (approved):

PR #361, opened 2023-01-16
Refactor/cleanup - topology fetch

mmajcher:


src/lib/topology/topology.py
```diff
@@ -0,0 +1,67 @@
...
+
+    topology_by_service: Dict[GCPService, Iterable[Entity]] = {}
+
+    for service, task in topology_tasks_by_service.items():
+        topology_by_service[service] = await task
+
+    return topology_by_service
+
+
+def choose_services_for_topology_fetch(
```
joaquinfilipic-dynatrace, line 28: It's my duty to state my opinion: we are adding another iteration with more lines of code for a very simple task.
But again, it's subjective and I get the approach :), I approve anyway.

joaquinfilipic-dynatrace (approved):

PR #365, opened 2023-01-19
PCLOUDS-1705: Add release version

joaquinfilipic-dynatrace:


src/lib/utilities.py
```diff
@@ -22,6 +22,8 @@
 
 K8S_CONTAINER_NAME_PREFIX = "dynatrace-gcp-function"
 
+RELEASE_VERSION = os.environ.get("TRAVIS_TAG", "")
```
mmajcher, line 25: I think the code should not know about pipeline technology we have, this is minor but would prefer to have env var also named "RELEASE_VERSION"

src/lib/utilities.py
```diff
@@ -72,6 +74,10 @@ def is_deployment_running_inside_cloud_function():
     return K8S_CONTAINER_NAME_PREFIX not in HOSTNAME
 
 
+def get_release_version():
```
mmajcher, line 77: Also we have config.py for getting configuration/env vars :P

src/lib/configuration/config.py
```diff
@@ -36,3 +36,6 @@ def dynatrace_log_ingest_url_secret_name():
 def keep_refreshing_extensions_config():
     return os.environ.get("KEEP_REFRESHING_EXTENSIONS_CONFIG", "TRUE").upper() in ["TRUE", "YES"]
 
+
+def release_tag():
+    return os.environ.get("RELEASE_TAG", "1.x.x")
```
mmajcher, line 41: "1.x.x" -> "version value not provided"

mmajcher (approved):

PR #369, opened 2023-01-27
Pclouds 1707 api latency statistics

NematulloKozimov:


src/lib/sfm/api_call_latency.py
```diff
@@ -0,0 +1,19 @@
...
+from collections import defaultdict
+
+
+class ApiCallLatency:
+    _value = defaultdict(list)
+
+    @staticmethod
+    def update(api_url, time):
+        if api_url in ApiCallLatency._value:
+            ApiCallLatency._value[api_url].append(time)
```
mmajcher, line 11: out of this 4 lines, I think only 11 is needed if you use defaultdict. also in "print_statistics" when you reset _value it should be also defaultdict

src/lib/sfm/api_call_latency.py
```diff
@@ -0,0 +1,19 @@
...
+        else:
+            ApiCallLatency._value[api_url] = [time]
+
+    @staticmethod
+    def print_statistics(context: LoggingContext):
+        for api_url, times in ApiCallLatency._value.items():
+            context.log(f"Url: {api_url}, min: {min(times):.3}s, avg: {sum(times) / len(times):.3}s, max: {max(times):.3}s")
```
mmajcher, line 18: Maybe lets write "Call latency for url: X, (...) ", also I think it can be one log line, usually it will be not that interesting, when you are really interested you will dig in

NematulloKozimov:
Screenshot 2023-02-20 at 09 00 40

mmajcher (approved):

PR #383, opened 2023-02-28
PCLOUDS-1872: AG Auth token

joaquinfilipic-dynatrace:


nikita-paskal-dynatrace (approved):

dtwwolinski:
In 'deploy-helm' we have a check for connection with given Pass token to check if the token is a valid one.
`check_dynatrace_docker_login`
Can we also have it with auth token?

dtwwolinski (changes requested):
As above.

PR #384, opened 2023-02-28
PCLOUDS-1710 Removed sending limit and added configurable buffer

NematulloKozimov:


nikita-paskal-dynatrace (approved):

src/lib/configuration/config.py
```diff
@@ -57,3 +61,6 @@ def gcp_monitoring_url():
     return os.environ.get("GCP_MONITORING_URL", "https://monitoring.googleapis.com/v3")
 
 
+
```
nikita-paskal-dynatrace, line 64: Do you really need all this empty lines?

nikita-paskal-dynatrace (approved):

PR #386, opened 2023-03-06
Deleted AG statefulSet file and related references.

joaquinfilipic-dynatrace:


scripts/deploy-helm.sh
```diff
@@ -51,41 +51,6 @@ check_dynatrace_log_ingest_url() {
   fi
 }
 
-check_activegate_state() {
```
joaquinfilipic-dynatrace, line 54: We are already checking connection (even sending a test log line through it) to provided ingestLogsUrl

nikita-paskal-dynatrace (approved):

dtwwolinski (approved):

PR #399, opened 2023-03-28
- added decode of token to fix the integrations tests failure.

dtwwolinski:


dtwwolinski:
My bad it turned out that I had an old version of PyJWT. Deleting the PR.

PR #406, opened 2023-04-19
fix cron tests

olegkozel:
cron tests are failing, because at first, they are running gcp monitor only with logs container


nikita-paskal-dynatrace (approved):
Good job :)

joaquinfilipic-dynatrace (approved):

PR #407, opened 2023-04-28
PCLOUDS-2225_GCP-permissions

nikita-paskal-dynatrace:
Removed unnecessary permissions during the deployment.


joaquinfilipic-dynatrace:
As far as I know, we are no longer using container.services, replicaSets or statefulSets at all now. So we might as well delete all related permissions to those resources too.
But anyway, worth to confirm. Leaving "get" permissions is not a big deal.

PR #413, opened 2023-06-19
PCLOUDS-2431_vulnerability-fix

nikita-paskal-dynatrace:
Vulnerability-fix


dtwwolinski (commented):
I think, you should use 3.8 python version instead of 3.9.

PR #419, opened 2023-07-18
PCLOUDS-2592 Initial Commit

szymon-nagel-dt:
W.I.P


src/lib/dt_extensions/dt_extensions.py
```diff
@@ -22,7 +25,10 @@ async def prepare_services_config_for_next_polling(current_services: List[GCPSer
             extensions_fetch_result = await extensions_fetch(gcp_session, dt_session, token)
             if not extensions_fetch_result:
                 raise Exception('Extension fetch failed')
-
+            
+            if config.metric_autodiscovery():
```
nikita-paskal-dynatrace, line 29: return await enrich_services_autodiscovery(extensions_fetch_result,gcp_session,dt_session,token) if config.metric_autodiscovery()

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+                "unit": self.unit,
+                "tags": self.tags,
+                "metricProperties": self.metric_properties,
+                "dimensions": self.dimensions
+            }
+        })
+
+
+def unit_caster(unit: str) -> str:
+    if unit in ["1","count","Count", "{operation}","{packet}" ,"{packets}" ,"{request}" ,"{port}" ,"{connection}", "{devices}" ,"{errors}" ,"{cpu}", "{query}", "{inode}"]:
```
nikita-paskal-dynatrace, line 63: case ?

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+        return "PerSecond"
+    elif unit == "frames/seconds":
+        return "PerSecond"
+    elif unit == "s{CPU}":
+        return "Second"
+    elif unit == "s{uptime}":
+        return "Second"
+    elif unit == "{dBm}":
+        return "DecibelMilliWatt"
+    else:
```
nikita-paskal-dynatrace, line 103: return "Unspecified"

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+    if metrickind == "DELTA" and valuetype != "DISTRIBUTION":
+        return "count,delta"
+    else:
+        return ""
+
+    
+def metric_name_parse(metric_name: str,metric_type: str) -> str:
+   # . -> _
+   # / -> .
+   metric_name.replace(r'\.','_')
```
nikita-paskal-dynatrace, line 121: return metric_name.replace(r'\.','_').replace(r'/','.')

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+    headers = {
+    "Accept": "application/json",
+    "Authorization": f"Bearer {token}"}
+    url = f"https://monitoring.googleapis.com/v3/projects/{projectID}/metricDescriptors"
+    resp = await gcp_session.request('GET', url=url, headers=headers)
+    page = await resp.json()
+
+    metric_yaml = [
+        {
+            "value": x.get("type"),
```
nikita-paskal-dynatrace, line 165: x?

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+                    "unit":         unit_caster(x.get("unit"))
+            },
+            "dimensions": [ 
+                {
+                "key":      dimension.get("key"), 
+                "value":    "label:metric.labels." + dimension.get("key"),
+                } for dimension in x.get("labels") or []],
+            "monitored_resources_types": x.get("monitoredResourceTypes",[])
+        } for x in page["metricDescriptors"]]
+
```
nikita-paskal-dynatrace, line 186: so many lines

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+    #Metadata Metrics
+    #ret = await _sent_metadata_to_dt(dynatrace_url[0],dynatrace_access_key,dt_session,metric_metadata_list)
+    
+    return gcp_services_list
+
+
+
+
+def main():
+    os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = r"C:\Users\szymon.nagel\Downloads\dynatrace-gcp-extension-0618ff31e584.json"
```
nikita-paskal-dynatrace, line 232: ???

src/run_autodiscovery.py
mmajcher: I think it should be a dictionary, you look up by the key

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+
+logging_context = LoggingContext(None)
+projectID="dynatrace-gcp-extension"
+discovered_resource_type="cloud_function"
+
+
+
+@dataclass(frozen=True)
+class MetricMetadata:
+    """Describes singular GCP service to ingest data from."""
```
mmajcher, line 22: If it describes singular GCP service, why it is named MetricMetadata? 
It looks more like it describes single metric?

src/lib/dt_extensions/dt_extensions.py
mmajcher: I don't think we should change it like this, autodiscovery is kind of a additional feature and not really relevant for main return. Just because we can squash everything to one line does not mean we should!

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+from dataclasses import dataclass
+from lib.clientsession_provider import init_gcp_client_session
+from lib.context import LoggingContext, MetricsContext, get_should_require_valid_certificate
+from lib.credentials import create_token, fetch_dynatrace_api_key, fetch_dynatrace_url
+from lib.metrics import Metric
+from typing import List, Text, Any, Dict
+from lib.configuration import config
+
+logging_context = LoggingContext(None)
+projectID="dynatrace-gcp-extension"
```
mmajcher, line 15: Please de-hardcode

mmajcher:
Please do not merge it to master for now, please create some branch for all autodiscovery stuff and each new branch (per ticket) merge to that autodiscvoery branch

src/run_autodiscovery.py
mmajcher: And also I would move all those dictionaries to separate file so it is easier to read the code

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+        return "gauge"
+    if metrickind == "DELTA" and valuetype == "DISTRIBUTION":
+        return "gauge"
+    if metrickind == "DELTA" and valuetype != "DISTRIBUTION":
+        return "count,delta"
+    else:
+        return ""
+
+    
+def metric_name_parse(metric_name: str,metric_type: str) -> str:
```
mmajcher, line 118: parse = read 
you are kinda transforming, not really reading? Please make it more descriptive. I think it is "change_metric_key_from_gcp_format_to_dt_format" (I am not sure though if this is the functionality)

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
+import asyncio
```
mmajcher, line 1: Please move this file to src/lib/autodiscovery/autodiscovery.py

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+    unit: Text
+    tags: List[Text]
+    dimensions: List[Dict[Text,Text]]
+    metric_properties: Dict[Text,Any]
+    schema_id: Text
+    scope: Text
+
+
+    def __init__(self, **kwargs):
+        object.__setattr__(self, "displayName", kwargs.get("displayName", ""))
```
mmajcher, line 37: I think python constructor should look like this
image

I know we have this weird setattr in some places, we have to remove them :D But for new code please make it clean

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+        headers={
+            "Authorization": f"Api-Token {dynatrace_access_key}",
+            "Content-Type": "application/json"
+        },
+        data=data_input,
+        verify_ssl=get_should_require_valid_certificate()
+    )
+
+    if response.status != 200:
+        print("Error During sending metadata meterics")
```
mmajcher, line 149: only use context.log (context.error, context.exception)

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+    page = await resp.json()
+
+    metric_yaml = [
+        {
+            "value": x.get("type"),
+            "key":  "cloud.gcp." + metric_name_parse(x.get("type"),type_metrickind_caster(x.get("metricKind"),x.get("valueType"))),
+            "displayName":  x.get("displayName"),
+            "name": x.get("displayName"),
+            "description": x.get("description"),
+            "type": type_metrickind_caster(x.get("metricKind"),x.get("valueType")),
```
mmajcher, line 170: Can we have method more descriptive? It is casting from GCP kind to DT kind, right? Can you put that in the name?

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+
+async def get_metric_descriptors(gcp_session,token):
+    headers = {
+    "Accept": "application/json",
+    "Authorization": f"Bearer {token}"}
+    url = f"https://monitoring.googleapis.com/v3/projects/{projectID}/metricDescriptors"
+    resp = await gcp_session.request('GET', url=url, headers=headers)
+    page = await resp.json()
+
+    metric_yaml = [
```
mmajcher, line 163: This array of dictionaries that you create here is not metric_yaml. At least not yaml :D
This is the format for metric from extensions yaml?
Then its more like "fetch_metric_definition"?

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+async def enrich_services_autodiscovery(gcp_services_list,gcp_session,dt_session,token):
+    print("Start AutoDiscovery")
+
+
+    bucket_gcp_services =  list(filter(lambda x: discovered_resource_type in x.name, gcp_services_list[0]))
+    existing_metric_list = [metric for service in bucket_gcp_services for metric in service.metrics]
+    all_metrics_list = await get_metric_descriptors(gcp_session,token)
+
+
+    missing_metrics_list = [metric for metric in all_metrics_list if metric[0].google_metric not in [existing_metric.google_metric for existing_metric in existing_metric_list]]
```
mmajcher, line 205: This is kinda packed, maybe lets try to extract some additional variables, then it is easier to read:
gcp_keys_from_metric_list = XXX
And then list comprehension, at least you will not nest it, there is no good reason to nest here

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+from lib.clientsession_provider import init_gcp_client_session
+from lib.context import LoggingContext, MetricsContext, get_should_require_valid_certificate
+from lib.credentials import create_token, fetch_dynatrace_api_key, fetch_dynatrace_url
+from lib.metrics import Metric
+from typing import List, Text, Any, Dict
+from lib.configuration import config
+
+logging_context = LoggingContext(None)
+projectID="dynatrace-gcp-extension"
+discovered_resource_type="cloud_function"
```
mmajcher, line 16: Did you discover this resource type? It looks like you put it here as string yourself :P 
So it is more like "resource_type_for_autodiscovery"

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+                } for dimension in x.get("labels") or []],
+            "monitored_resources_types": x.get("monitoredResourceTypes",[])
+        } for x in page["metricDescriptors"]]
+    
+    
+
+
+
+    metric_list = [(Metric(**x), x.get("monitored_resources_types"),x) for x in metric_yaml if x.get("gcpOptions").get("valueType").upper() != "STRING" ]
+    discovered_resource_type_list = list(filter(lambda x: discovered_resource_type in x[1], metric_list))
```
mmajcher, line 192: So you're like fetching all metric descriptors and then throwing most of them out to only have for function, right? 
1. How big is the response, how many metric descriptors, is there paging? 
2. Is this possible to add some filter to query so that you only fetch descriptors for lambda?

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+    
+    return gcp_services_list
+
+
+
+
+def main():
+    os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = r"C:\Users\szymon.nagel\Downloads\dynatrace-gcp-extension-0618ff31e584.json"
+    logging_context.log("Dynatrace GCP Metrics autodiscovery startup")
+    asyncio.run(init_session())
```
mmajcher, line 234: Why are you running new loop in here???

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+                service.metrics.extend([metric[0] for metric in missing_metrics_list])
+    
+
+    print(f"Adding Metadata to metrics")
+    dynatrace_url = await fetch_dynatrace_url(gcp_session, config.project_id(), token),
+    dynatrace_access_key=await fetch_dynatrace_api_key(gcp_session, config.project_id(), token),
+
+    metric_metadata_list = [MetricMetadata(**(x[2])) for x in missing_metrics_list]
+    #Metadata Metrics
+    #ret = await _sent_metadata_to_dt(dynatrace_url[0],dynatrace_access_key,dt_session,metric_metadata_list)
```
mmajcher, line 224: How long does it take? Maybe we should remember in memoery which metadata we sent, maybe even check if something changed. But sending metedata each time probably unnecessary
UNLESS we would be able to send metadata inside mint ingest line? Then it would be more data sent but less code here, for first iteration I would prefer this solution if possible

src/run_autodiscovery.py
mmajcher: Ahhh, you were trying to run the file yourself in dev approach, right?
Thats fine! Maybe a unit tests for this would be better, and actually can stay as useful unit test for future

mmajcher: Btw, before you merge it, please add unit tests

src/lib/dt_extensions/dt_extensions.py
mmajcher, line 29: And other thing is, I am not sure that dt_extensions should be running autodiscovery. Its like on different level.
You (lets say some main place) fetch dt_extensions, then you run autodiscovery and add to result of dt_extensions.

mmajcher: So I would move this enrich one level higher (calling method)

src/run_autodiscovery.py
```diff
@@ -0,0 +1,237 @@
...
+        return "Second"
+    elif unit == "s{uptime}":
+        return "Second"
+    elif unit == "{dBm}":
+        return "DecibelMilliWatt"
+    else:
+        return "Unspecified"
+    
+
+def type_metrickind_caster(metrickind: str, valuetype: str) -> str:
```
nikita-paskal-dynatrace, line 107: > def type_metrickind_caster(metrickind: str, valuetype: str) -> str:
    if metrickind == "GAUGE":
        return "gauge"
    if metrickind == "DELTA":
        if valuetype == "DISTRIBUTION":
            return "gauge"
        return "count,delta"

    return ""

PR #426, opened 2023-08-10
Pclouds autodiscovery fix

szymon-nagel-dt:
Fixing feature sets


src/lib/metrics.py
```diff
@@ -173,12 +173,14 @@ class GCPService:
     metrics = List[Metric]
     monitoring_filter: Text
     activation: Dict[Text, Any]
+    is_configured: bool
```
joaquinfilipic-dynatrace, line 176: I would change this name. Is configured... where? Or to do what? This means that it's enabled from feature sets, right?

PR #428, opened 2023-08-16
Added exclusion param for projects

joaquinfilipic-dynatrace:


src/main.py
```diff
@@ -126,6 +126,8 @@ async def query_metrics(execution_id: Optional[str], services: Optional[List[GCP
             disabled_projects, disabled_apis_by_project_id = \
                 await get_disabled_projects_and_disabled_apis_by_project_id(context, projects_ids)
 
+        disabled_projects.extend(filter(None, config.excluded_projects().split(',')))
```
dtwwolinski, line 129: Imho it should be set

dtwwolinski (changes requested):

dtwwolinski (approved):

nikita-paskal-dynatrace (approved):

PR #434, opened 2023-09-06
Pclouds 2665 autodiscovery for all extensions that we have

szymon-nagel-dt:


src/lib/autodiscovery/autodiscovery.py
```diff
@@ -29,10 +30,28 @@
 
 AutodiscoveryResult = NamedTuple(
     "AutodiscoveryResult",
-    [("enriched_services", List[GCPService]), ("discovered_metric_list", Dict[str, Any])],
+    [
+        ("autodiscovered_resources_to_metrics", Dict[str, List[Metric]]),
+        ("discovered_metric_list", Dict[str, Any]),
```
nikita-paskal-dynatrace, line 35: shouldn't it be metric`S`

src/lib/autodiscovery/autodiscovery.py
```diff
@@ -71,7 +125,14 @@ async def run_fetch_metric_descriptors(
                 metric_descriptor = GCPMetricDescriptor.create(**descriptor, project_id=project_id)
                 if (
                     metric_descriptor.gcpOptions.valueType.upper() != "STRING"
-                    and discovered_resource_type in metric_descriptor.monitored_resources_types
+                    and len(metric_descriptor.monitored_resources_types) == 1
```
nikita-paskal-dynatrace, line 128: Those multiple lines of logic could be good candidate for function

src/lib/autodiscovery/autodiscovery.py
```diff
@@ -90,36 +151,59 @@ async def run_fetch_metric_descriptors(
...
-    for metric in metrics:
-        if metric.dynatrace_name not in previously_discovered_metrics:
-            metrics_metadata.append(
-                MetadataIngestLine(
-                    metric_name=metric.dynatrace_name,
-                    metric_type=metric.dynatrace_metric_type,
-                    metric_display_name=metric.name,
-                    metric_description=metric.description,
-                    metric_unit=metric.unit,
+    for _, metric_list in metrics.items():
```
nikita-paskal-dynatrace, line 159: metric`s`_list

src/lib/autodiscovery/autodiscovery.py
```diff
@@ -140,60 +245,56 @@ async def run_autodiscovery(
...
+        monitored_resource = descriptor.monitored_resources_types[0]
+        if descriptor.value not in existing_resources_to_metrics[monitored_resource]:
+            autodiscovery_resources_to_metrics.setdefault(monitored_resource, []).append(
+                Metric(**(asdict(descriptor)), autodiscovered_metric=True, project_ids=project_ids)
+            )
 
-    missing_metrics_list = []
+    for resource_name, metric_list in autodiscovery_resources_to_metrics.items():
+        logging_context.log(
+            f"For resource: [{resource_name}] discovered {len(metric_list)} metrics"
```
nikita-paskal-dynatrace, line 272: What about: Discovered {len(metric_list)} metrics for [{resource_name}] resource.

src/lib/autodiscovery/autodiscovery_manager.py
```diff
@@ -1,76 +1,148 @@
...
+                if any(autodiscovery_result.autodiscovered_resources_to_metrics):
+                    autodiscovery_service.set_metrics(
+                        autodiscovery_result.autodiscovered_resources_to_metrics,
+                        resources_to_discovery,
+                        autodiscovery_result.resource_dimensions,
+                    )
+                    return autodiscovery_service
+
+            self.logging_context.log(
+                "No resources find to autodiscovery. Please ensure that you have either added them in the autodiscoveryResourcesYaml file or verified that the necessary extensions are enabled."
```
nikita-paskal-dynatrace, line 146: No resources been found by autodiscovery. Please`,` ensure you either

src/lib/autodiscovery/autodiscovery.py
```diff
@@ -131,7 +215,28 @@ async def get_metric_descriptors(
...
+
+
+async def get_existing_metrics(
+    autodiscovery_resources: Dict[str, AutodiscoveryResourceLinking],
+) -> Dict[str, Set[str]]:
+    resources_to_metrics = {
+        resource_name: set() for resource_name in autodiscovery_resources.keys()
+    }
+
+    for resource_name, service_linking in autodiscovery_resources.items():
```
nikita-paskal-dynatrace, line 228: At some moment I lost inside the level of the nested loops

k8s/helm-chart/dynatrace-gcp-monitor/values.yaml
```diff
@@ -124,9 +124,162 @@ metricResources:
     cpu: "1.25"
     ephemeral-storage: "2Gi"
 # metricAutodiscovery: If true additional metrics not listed in extensions will be automatically detected and monitored
-metricAutodiscovery: "true"
+metricAutodiscovery: "false"
 # Metrics autodiscovery polling interval in minutes.
 autodiscoveryQueryInterval: 60
+# configuration file for autodiscovery
+autodiscoveryResourcesYaml: |
```
joaquinfilipic-dynatrace, line 131: Could we change the approach of configuring the desired autodiscovered services. Customer can put a simple comma-separated list (or something like that) of services directly, assuming he knows them from Google (not to overload the values.yaml).

k8s/helm-chart/dynatrace-gcp-monitor/values.yaml
```diff
@@ -124,9 +124,162 @@ metricResources:
...
+    # - cloudsql_database
+    # - cloudsql_instance_database
+    # Google Virtual Private Cloud
+    # - vpc_access_connector
+    # Google Network Topology
+    # - gce_zone_network_health
+
+# configuration file for metrics that contains names bellow that shoud not be included
+# in autodiscovery process
+autodiscoveryBlockListYaml: |
```
joaquinfilipic-dynatrace, line 268: Same thing with block_list. Also, we could maybe avoid passing them as parameters through all the methods, because it's a long flow. If it's only some filtering-out list, it will be easier to read all methods if we remove those params and just get the list when needed using config.py.

k8s/helm-chart/dynatrace-gcp-monitor/values.yaml
```diff
@@ -124,9 +124,162 @@ metricResources:
...
+    # - spanner_instance
+    # Google Cloud SQL
+    # - cloudsql_database
+    # - cloudsql_instance_database
+    # Google Virtual Private Cloud
+    # - vpc_access_connector
+    # Google Network Topology
+    # - gce_zone_network_health
+
+# configuration file for metrics that contains names bellow that shoud not be included
```
joaquinfilipic-dynatrace, line 266: Check redaction. Eg. : "... metrics containing the following names will not be included."

src/lib/autodiscovery/autodiscovery.py
```diff
@@ -131,7 +215,28 @@ async def get_metric_descriptors(
...
+                    metric_name_list.append(metric.google_metric)
+            for service in service_linking.disabled_services_for_resource:
+                for metric in service.metrics:
+                    metric_name_list.append(metric.google_metric)
+        resources_to_metrics[resource_name] = set(metric_name_list)
+
+    return resources_to_metrics
 
 
 async def run_autodiscovery(
```
joaquinfilipic-dynatrace, line 242: Could we add some short comments on main methods, like this one, to say what is being done? Like in fetch_ingest_lines_task(), in main.py. This is an extra, we should do this not only in this PR 😄

src/lib/autodiscovery/autodiscovery.py
```diff
@@ -53,8 +72,43 @@ async def get_project_ids(
...
+
+    return resource_dimensions
+
+
 async def run_fetch_metric_descriptors(
-    gcp_session: ClientSession, token: str, project_id: str
+    gcp_session: ClientSession,
+    token: str,
+    project_id: str,
+    resources_to_autodiscovery: Set[str],
```
joaquinfilipic-dynatrace, line 110: resources_to_autodiscover?

src/lib/autodiscovery/autodiscovery.py
```diff
@@ -90,36 +151,59 @@ async def run_fetch_metric_descriptors(
...
-) -> Dict[GCPMetricDescriptor, List[str]]:
+    metric_context: MetricsContext,
+    gcp_session: ClientSession,
+    token: str,
+    autodiscovery_resources_to_services: Dict[str, AutodiscoveryResourceLinking],
+    autodiscovery_metric_block_list: List[str],
+) -> Tuple[Dict[GCPMetricDescriptor, List[str]], Dict[str, List[Dimension]]]:
     project_ids = await get_project_ids(metric_context, gcp_session, token)
 
+    resources_to_autodiscovery = {
```
joaquinfilipic-dynatrace, line 188: Idem, naming

src/lib/autodiscovery/autodiscovery.py
joaquinfilipic-dynatrace: True. It's worth to check if we can reduce all the nesting (other places too). Maybe it could be simplified.

src/lib/autodiscovery/autodiscovery_manager.py
```diff
@@ -1,76 +1,148 @@
...
+                                service_name=service["service_name"],
+                                feature_set_name=service["feature_set"],
+                            )
+                        )
+        return {
+            resource_name: list(set(service_stub))
+            for resource_name, service_stub in resource_mapping.items()
+        }
+
+    async def check_resources_to_autodiscovery(
```
joaquinfilipic-dynatrace, line 76: to_autodiscover?

src/lib/autodiscovery/autodiscovery_manager.py
```diff
@@ -1,76 +1,148 @@
...
+
+        for resource in self.autodiscovery_config["searched_resources"]:
+            if resource in self.autodiscovery_resource_mapping:
+                possible_linking = []
+                disabled_linking = []
+
+                for required_service in self.autodiscovery_resource_mapping[resource]:
+                    if required_service in enabled_services_identifiers:
+                        possible_linking.append(enabled_services_identifiers[required_service])
+                    if required_service in disabled_services_identifiers:
```
joaquinfilipic-dynatrace, line 101: Consider reducing noise in the logic too. E.g., from line 90, we don't need to ask again in 101, we know it will be disabled if it's not enabled (unless I am missing something). Stuff like that.

src/lib/autodiscovery/autodiscovery_manager.py
```diff
@@ -1,76 +1,148 @@
...
+                        disabled_services_for_resource=disabled_linking,
+                    )
+
+                else:
+                    self.logging_context.log(
+                        f"Can't add resource {resource}. Make sure if extension {self.autodiscovery_resource_mapping[resource][0].extension_name} with proper services is enabled."
+                    )
+
+            else:
+                prepared_resources[resource] = None
```
joaquinfilipic-dynatrace, line 116: What does this mean? If the resource is not in the mapping, you won't be looking for metrics from it? What about resources which we are not considering in any extensions?
If this is on purpose and it's to be extended, no problem. Next PR :)

src/lib/autodiscovery/autodiscovery_manager.py
```diff
@@ -1,76 +1,148 @@
...
-                services, self.last_autodiscovered_metric_list_names
+        self.autodiscovery_enabled = True
+
+        self._load_config()
+
+    def _load_config(self):
+        try:
+            self.autodiscovery_config = read_autodiscovery_config_yaml()["autodicovery_config"]
+
+            with open("./lib/autodiscovery/config/autodiscovery-mapping.yaml", "r") as file_mapping:
```
joaquinfilipic-dynatrace, line 44: How will the maintenance of the autodiscovery-mapping look like? What if we change our extensions, or add more? Will we have to change this too?

src/lib/autodiscovery/autodiscovery_manager.py
```diff
@@ -1,76 +1,148 @@
...
-        self.autodiscovery_task = asyncio.create_task(
-            enrich_services_with_autodiscovery_metrics(
-                services, self.last_autodiscovered_metric_list_names
+        self.autodiscovery_enabled = True
+
+        self._load_config()
+
+    def _load_config(self):
+        try:
+            self.autodiscovery_config = read_autodiscovery_config_yaml()["autodicovery_config"]
```
joaquinfilipic-dynatrace, line 42: How does autodicovery_config file differ from the selected resources in values.yaml?

src/lib/autodiscovery/autodiscovery.py
```diff
@@ -202,12 +303,23 @@ async def enrich_services_with_autodiscovery_metrics(
...
+                previously_discovered_metrics,
+                autodiscovery_resources,
+                autodiscovery_metric_block_list,
             )
 
             return autodiscovery_fetch_result
     except Exception as e:
         logging_context.error(
-            f"Failed to prepare autodiscovery new services metrics, will reuse previous metrics list; {str(e)}"
+            f"Failed to prepare autodiscovery new services metrics, will reuse previous metrics list; "
```
joaquinfilipic-dynatrace, line 318: "... autodiscovery for new services"? "Of new services"?

src/lib/autodiscovery/autodiscovery_manager.py
```diff
@@ -1,76 +1,148 @@
...
+            else:
+                prepared_resources[resource] = None
+
+        return prepared_resources
+
+    async def get_autodiscovery_service(
+        self, services: List[GCPService]
+    ) -> Optional[AutodiscoveryGCPService]:
+        if self.autodiscovery_enabled:
+            resources_to_discovery = await self.check_resources_to_autodiscovery(services)
```
joaquinfilipic-dynatrace, line 124: to_discover

src/lib/autodiscovery/autodiscovery_task_executor.py
```diff
@@ -0,0 +1,108 @@
+import asyncio
+from datetime import datetime, timedelta
+from typing import Dict, List, Optional
+
+from lib.autodiscovery.autodiscovery_manager import AutodiscoveryManager
+from lib.configuration import config
+from lib.context import LoggingContext
+from lib.metrics import AutodiscoveryGCPService, GCPService
+
+AUTODISCOVERY_QUERY_INTERVAL_SEC = config.get_autodiscovery_querry_interval() * 60
```
joaquinfilipic-dynatrace, line 10: Besides this var not being used, isn't the autodiscovery also depending on the main loop's query interval (inside run_docker's loop)? Maybe not, just asking.

src/lib/autodiscovery/autodiscovery_task_executor.py
```diff
@@ -0,0 +1,108 @@
...
+    ):
+        new_extension_versions_hash = hash(tuple(sorted(new_extension_versions.items())))
+        delta = datetime.now() - self.time_since_last_autodiscovery
+
+        if self.autodiscovery_task is not None and self.autodiscovery_task.done():
+            await self._handle_autodiscovery_task_result()
+
+        if (
+            new_extension_versions_hash != self.autodiscovered_extension_versions_hash
+            or delta > self.query_interval
```
joaquinfilipic-dynatrace, line 100: Does it make sense to ask this if the hash is already equal? I mean, there's an OR, if delta is > than the interval but the hash didn't change, why to run the following task?

src/lib/autodiscovery/autodiscovery_task_executor.py
```diff
@@ -0,0 +1,108 @@
...
+                self.autodiscovered_cached_service = ad_service_result
+            self.autodiscovery_task = None
+
+    async def get_cached_or_refreshed_metrics(
+        self, services: List[GCPService], new_extension_versions: Dict[str, str]
+    ):
+        new_extension_versions_hash = hash(tuple(sorted(new_extension_versions.items())))
+        delta = datetime.now() - self.time_since_last_autodiscovery
+
+        if self.autodiscovery_task is not None and self.autodiscovery_task.done():
```
joaquinfilipic-dynatrace, line 95: Asking multiple times if self.autodiscovery_task.done() and repeating other code (isn't get_task_result ~= _handle_autodiscovery_task_result). We should simplify logic in this method.

src/lib/autodiscovery/autodiscovery_task_executor.py
```diff
@@ -0,0 +1,108 @@
...
+from lib.configuration import config
+from lib.context import LoggingContext
+from lib.metrics import AutodiscoveryGCPService, GCPService
+
+AUTODISCOVERY_QUERY_INTERVAL_SEC = config.get_autodiscovery_querry_interval() * 60
+
+logging_context = LoggingContext("AUTODISCOVERY_TASK")
+
+
+class AutodiscoveryTaskExecutor:
```
joaquinfilipic-dynatrace, line 15: Methods and logic could be simplified overall.

src/lib/autodiscovery/autodiscovery_task_executor.py
```diff
@@ -0,0 +1,108 @@
...
+        if (
+            new_extension_versions_hash != self.autodiscovered_extension_versions_hash
+            or delta > self.query_interval
+        ):
+            await self._refresh_autodiscovery_task(services, new_extension_versions_hash)
+
+        if self.autodiscovered_cached_service:
+            services.append(self.autodiscovered_cached_service)
+        else:
+            logging_context.log("No resources to autodiscovery will skip")
```
joaquinfilipic-dynatrace, line 107: Check redaction

src/lib/metrics.py
```diff
@@ -200,11 +210,66 @@ def __init__(self, **kwargs):
         monitoring_filter = VARIABLE_VAR_PATTERN.sub('', monitoring_filter)
         object.__setattr__(self, "monitoring_filter", monitoring_filter)
         object.__setattr__(self, "is_enabled",  kwargs.get("is_enabled", True))
+        object.__setattr__(self, "extension_name",  kwargs.get("extension_name", "Unknown Extension"))
 
     def __hash__(self):
         return hash((self.name, self.technology_name, self.feature_set, self.monitoring_filter))
 
 
+class AutodiscoveryGCPService(GCPService):
```
joaquinfilipic-dynatrace, line 219: Think if all the internal relations are really necessary or could be avoided. Here, this has many Dicts. We could come back later to this, maybe there's something to do to reduce it, maybe not. Just to double check

src/lib/autodiscovery/autodiscovery_task_executor.py
```diff
@@ -0,0 +1,108 @@
...
+            self.query_interval = timedelta(minutes=60)
+
+        self.autodiscovery_task = asyncio.create_task(
+            self.autodiscovery_manager.get_autodiscovery_service(services)
+        )
+        self.time_since_last_autodiscovery = datetime.now()
+
+    async def get_task_result(self):
+        if self.autodiscovery_task is not None:
+            ad_service_result = await self.autodiscovery_task
```
joaquinfilipic-dynatrace, line 63: add_service_result?

src/lib/autodiscovery/autodiscovery_task_executor.py
```diff
@@ -0,0 +1,108 @@
...
+        self.autodiscovery_task = asyncio.create_task(
+            self.autodiscovery_manager.get_autodiscovery_service(services)
+        )
+
+        self.autodiscovered_extension_versions_hash = new_extension_versions_hash
+        self.time_since_last_autodiscovery = datetime.now()
+
+    async def _handle_autodiscovery_task_result(self):
+        if self.autodiscovery_task:
+            ad_service_result = await self.autodiscovery_task
```
joaquinfilipic-dynatrace, line 84: Same here, add

k8s/helm-chart/dynatrace-gcp-monitor/values.yaml
szymon-nagel-dt: moved to autodiscovery-values.yaml

src/lib/autodiscovery/autodiscovery_manager.py
szymon-nagel-dt: This is the exact file - we have to put those from values.yaml inside kubernetes cluster

src/lib/autodiscovery/autodiscovery_task_executor.py
szymon-nagel-dt: refactored

PR #436, opened 2023-09-15
feat: added excluded_projects_by_prefix/substring capability

eandresr:
In addition to the full project_id filter, now this is able to filter by string contained into the project_id. Useful when yo have your GCP envs named with certain maner such as <provider>-<env>-<project>-<random-id> and you have one Dynatrace cluster for each env/proj/***


eandresr:
Implemented both python and Helm parts. Also added the comments into the values.yaml to clarify "why"

dtwwolinski (approved):

mmajcher:
Hi @eandresr, could you contact me at marek.majcher at dynatrace.com? Thanks :)

PR #474, opened 2024-01-24
Pclouds 3518 adjust tests

NematulloKozimov:
- Used asyncio for http requests
- Introduced multiprocessing
- Changed architecture (each worker does the whole process: pull->process->send_logs->send_ack_ids)
- Adjusted Self monitoring
- Got rid of Threads


src/lib/logs/gcp_client.py
```diff
@@ -0,0 +1,83 @@
...
+        self.acknowledge_url = f"https://pubsub.googleapis.com/v1/{SUBSCRIPTION_PATH}:acknowledge"
+        self.headers = {"Authorization": f"Bearer {api_token}"}
+        json_body = {"maxMessages": PROCESSING_WORKER_PULL_REQUEST_MAX_MESSAGES}
+        json_data = json.dumps(json_body)
+        self.body_payload = json_data.encode("utf-8")
+
+    async def pull_messages(
+        self, logging_context: LoggingContext, gcp_session
+    ) -> Dict[str, List[Any]]:  # type: ignore
+        async with gcp_session.request(
```
emjot13-dynatrace, line 55: Hey, just a question - is timing out handled in this request? Do we know it won't hang forever?

src/lib/logs/log_integration_service.py
```diff
@@ -0,0 +1,117 @@
...
+        self.gcp_client = None
+        self.dynatrace_client = DynatraceClient()
+        self.sfm_queue = sfm_queue
+        self.log_push_semaphore = asyncio.Semaphore(NUMBER_OF_CONCURRENT_PUSH_COROUTINES)
+
+    async def update_gcp_client(self, logging_context: LoggingContext):
+        async with init_gcp_client_session() as gcp_session:
+            token = await create_token(logging_context, gcp_session)
+            if token is None:
+                raise Exception("Cannot start pub/sub pulling - Failed to fetch token")
```
emjot13-dynatrace, line 34: Is it possible to make this Exception more informative - like why we failed to fetch the token?

src/lib/logs/log_integration_service.py
```diff
@@ -0,0 +1,117 @@
...
+
+            context.self_monitoring.processing_time_start = time.perf_counter()
+            for response in responses:
+                if not isinstance(response, Exception):
+                    for received_message in response.get("receivedMessages", []):  # type: ignore
+                        message_job = prepare_context_and_process_message(
+                            self.sfm_queue, received_message
+                        )
+                        if not message_job or message_job.bytes_size > REQUEST_BODY_MAX_SIZE - 2:
+                            ack_ids_to_send.append(received_message.get("ackId"))
```
emjot13-dynatrace, line 71: I think this for loop should be less complex. Could you consider for example creating an additional function to separate the logic?

src/lib/logs/log_integration_service.py
```diff
@@ -0,0 +1,117 @@
...
+        put_sfm_into_queue(context)
+
+        return ack_ids_to_send
+
+    async def push_ack_ids(self, ack_ids: List[str], logging_context: LoggingContext):
+        # request size limit is 524288, but we are not able to easily control size of created protobuf
+        # empiric test indicates that ack_ids have around 200-220 chars. We can safely assume that ack id is never longer
+        # than 256 chars, we split ack ids into chunks with no more than 2048 ack_id's
+        async with init_gcp_client_session() as gcp_session:
+            chunk_size = 2048
```
emjot13-dynatrace, line 110: Can this be a constant?

src/lib/logs/logs_processor.py
```diff
@@ -14,51 +14,141 @@
...
+    logs_for_next_batch: List[str] = []
+    ack_ids_for_next_batch: List[str] = []
+    logs_for_next_batch_total_len = 0
+    logs_for_next_batch_events_count = 0
+
+    log_entries = 0
+    batch_sfm_monitoring_list = []
+    for log_entry in logs:
+        new_batch_len = (
+            logs_for_next_batch_total_len + 2 + len(logs_for_next_batch) - 1
```
emjot13-dynatrace, line 73: The comment is great, but I thing that assigning a variable like "bracket_length = 2" would be even better 😄

src/lib/logs/logs_processor.py
```diff
@@ -14,51 +14,141 @@
...
+class LogBatch(NamedTuple):
+    serialized_batch: str
+    number_of_logs_in_batch: int
+    ack_ids: List[str]
+    size_batch_bytes: int
+    self_monitoring: LogSelfMonitoring
+
+
+
+def prepare_batches(logs: List[LogProcessingJob]) -> List[LogBatch]:
```
emjot13-dynatrace, line 61: As far as I can see this "batch preparation" consists of a few steps, so maybe could you consider splitting this into more functions?

src/lib/logs/logs_processor.py
```diff
@@ -67,20 +157,19 @@ def _prepare_context_and_process_message(sfm_queue: Queue, message: ReceivedMess
...
 
 
-def _process_message(context: LogsProcessingContext, message: PubsubMessage) -> Optional[LogProcessingJob]:
-    context.self_monitoring.processing_time_start = time.perf_counter()
-    data = message.data.decode("UTF-8")
-    # context.log(f"Data: {data}")
+def _process_message(
+    context: LogsProcessingContext, message: PubsubMessage, ack_id
+) -> Optional[LogProcessingJob]:
+    data = base64.b64decode(message.get("data"))
```
emjot13-dynatrace, line 163: Are we sure this message object will always have a "data" field?
If not, the data value will be set to None, which will throw an AttributeError on the next line.

src/run_docker.py
```diff
@@ -174,8 +176,11 @@ def main():
     if OPERATION_MODE == OperationMode.Metrics:
         asyncio.run(run_metrics_fetcher_forever())
     elif OPERATION_MODE == OperationMode.Logs:
-        LogsFastCheck(logging_context, instance_metadata).execute()
-        run_logs(logging_context, instance_metadata)
+        asyncio.run(LogsFastCheck(logging_context, instance_metadata).execute())
+        processes = [multiprocessing.Process(target=run_logs_wrapper, args=(logging_context, instance_metadata, i))
+                     for i in range(PARALLEL_PROCESSES)]
+        [process.start() for process in processes]
+        [process.join() for process in processes]
```
emjot13-dynatrace, line 183: I think it's worth considering to call `process.close() `for each process, so it releases resources immediately. 
Also just a small question out of curiosity, was there any particular reason why you used lists for this iteration 😄

PR #475, opened 2024-01-25
Pclouds 2800 gcp log forwarder perf improvements

NematulloKozimov:


k8s/helm-chart/dynatrace-gcp-monitor/values.yaml
```diff
@@ -96,8 +96,14 @@ logIngestRequestMaxSize: 1048576
 logIngestSendingWorkerExecutionPeriod: 60
 # Determines max age of forwarded log event. Should be the same or lower than on cluster
 logIngestEventMaxAgeSeconds: 86400
+# Number of parallel processes to run the whole log monitoring loop
+parallelProcesses: 1
 # Number of workers pulling logs from pubsub concurrently
 processingWorkers: 5
+# Number of concurrent coroutines to pull messages from pub/sub
+numberOfConcurrentMessagePullCoroutines: 10
+# Number of concurrent coroutines to push messaged to Dynatrace
```
joaquinfilipic-dynatrace, line 105: _messaged_, typo

src/dashboards/dynatrace-gcp-monitor-log-self-monitoring.json
```diff
@@ -258,6 +258,34 @@
           }
         }
       },
+      {
+        "title": "Pulling time",
```
joaquinfilipic-dynatrace, line 262: Nice ;)

src/lib/logs/log_forwarder.py
```diff
@@ -12,134 +12,75 @@
...
-    pull_request.subscription = subscription_path
+    log_integration_service = LogIntegrationService(sfm_queue)
+    await log_integration_service.update_gcp_client(logging_context)
+
+    tasks = []
+
+    update_token = asyncio.create_task(log_integration_service.keep_gcp_token_updated(logging_context))
+
+    tasks.append(update_token)
+    for worker_number in range(0, PROCESSING_WORKERS):
```
joaquinfilipic-dynatrace, line 55: So, I got confused. Do we still need this for? I believe based on tests, yes.
But we have individual sets of coroutines for pulling and pushing, this on top of it, and multiprocessing on top of everything??

k8s/helm-chart/dynatrace-gcp-monitor/templates/function-config.yaml
```diff
@@ -44,7 +44,10 @@ data:
   DYNATRACE_LOG_INGEST_REQUEST_MAX_SIZE: {{ .Values.logIngestRequestMaxSize | quote }}
   DYNATRACE_LOG_INGEST_SENDING_WORKER_EXECUTION_PERIOD: {{ .Values.logIngestSendingWorkerExecutionPeriod | quote }}
   DYNATRACE_TIMEOUT_SECONDS: {{ .Values.logIngestEventMaxAgeSeconds | quote }}
+  PARALLEL_PROCESSES: {{ .Values.parallelProcesses | quote }}
   PROCESSING_WORKERS: {{ .Values.processingWorkers | quote }}
```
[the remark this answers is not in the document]
joaquinfilipic-dynatrace: This will depend on comments below

src/lib/logs/log_forwarder.py
joaquinfilipic-dynatrace, line 55: So, we are having independent coroutines for pulling and pushing. More than one worker for each of them, and more than one process on top of everything, right?

joaquinfilipic-dynatrace: Could we change the name of PROCESSING_WORKERS? NUMBER_OF_CONCURRENT_something? E.g. NUMBER_OF_CONCURRENT_FORWARDING_LOOPS

src/lib/logs/log_forwarder.py
```diff
@@ -12,134 +12,75 @@
...
+        worker_task = asyncio.create_task(
+            pull_and_push_logs_forever(process_number, worker_number, log_integration_service)
+        )
+        tasks.append(worker_task)
+
+    sfm_task = asyncio.create_task(create_sfm_loop(sfm_queue, logging_context, instance_metadata))
+    tasks.append(sfm_task)
+
+    await asyncio.gather(*tasks)
```
joaquinfilipic-dynatrace, line 64: Out of not knowing. We're awaiting for tasks that run in 'while True' (create_sfm_loop, for instance), so why?

src/lib/logs/log_forwarder.py
```diff
@@ -12,134 +12,75 @@
...
+async def pull_and_push_logs_forever(
+    process_number: int, worker_number: int, log_integration_service: LogIntegrationService
+):
+    logging_context = LoggingContext(f"Process-{process_number}", f"Worker-{worker_number}")
     logging_context.log(f"Starting processing")
     while True:
         try:
-            perform_pull(worker_state, sfm_queue, subscriber_client, subscription_path, pull_request)
+            # Pull logs from pub/sub, process them and create batches out of them to send to Dynatrace
+            log_batches, ack_ids = await log_integration_service.perform_pull(logging_context)
```
joaquinfilipic-dynatrace, line 75: These ack_ids are only from the erroneous messages, as I understand (and we are adding later the ids from the sent messages). So we can also be more descriptive with the naming.

joaquinfilipic-dynatrace:
I would add also the diagram to the img folder. Or keep it somewhere, at least :D

src/lib/logs/log_forwarder.py
szymon-nagel-dt: If i am not mistaken we have to add await for corutines in order for them to schedule. Not having will result in warning message :  RuntimeWarning: Coroutine Was Never Awaited

src/lib/logs/gcp_client.py
```diff
@@ -0,0 +1,89 @@
...
+        self, logging_context: LoggingContext, gcp_session, update_gcp_client: Callable[[LoggingContext], None]
+    ) -> Dict[str, List[Any]]:  # type: ignore
+        async with gcp_session.request(
+            method="POST", url=self.pull_url, data=self.body_payload, headers=self.headers
+        ) as response:
+            response_json = await response.json()
+            resp_status = response.status
+
+            if resp_status == 401:
+                await update_gcp_client(logging_context)
```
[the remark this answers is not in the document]
joaquinfilipic-dynatrace: We are sending several pull_messages() tasks, but are we sending them all with the same token? Can they all fail and run update_gcp_client() many times? And we won't pull any messages in that loop. Anyway, just asking, not sure if that can happen.

joaquinfilipic-dynatrace:
Another thing; I might have missed it, though.
Are we doing something to reduce the whole pull-push rate if there is none or few data? Is that something to consider? If there are no logs, is it worth to add a (e.g. minute) sleep, not to be constantly working and creating requests?

src/lib/logs/gcp_client.py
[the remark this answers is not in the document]
NematulloKozimov: yes, multiple update of the token might happen.

NematulloKozimov:
> 

image

Looks like pub/sub only charges for bytes not for the requests

src/lib/logs/gcp_client.py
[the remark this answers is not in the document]
joaquinfilipic-dynatrace: If that can happen, it will happen every time we need to update it, right? 
So, each time the token expires, we update it NUMBER_OF_CONCURRENT_MESSAGE_PULL_COROUTINES times? 👀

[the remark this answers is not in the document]
joaquinfilipic-dynatrace: What if in log_integration_service, in the loop from line 55, we do the update once (making sure of this, extra variable or anything) if we see the specific response stating that the token expired?
Plus, we avoid sending update_gcp_client as param to pull_messages().

joaquinfilipic-dynatrace (dismissed):

joaquinfilipic-dynatrace (approved):

PR #476, opened 2024-01-25
PCLOUDS-2666 removed [AUTODISCOVERY] from metric name

szymon-nagel-dt:
removed [AUTODISCOVERY] from metric name
and added metadata.origin=autodiscovery


joaquinfilipic-dynatrace:
I assume it's tested 😛 
Can I check how it looks like? If you tried it on any tenant, the metrics with the dimension should appear, if I set the timeframe right.

joaquinfilipic-dynatrace (approved):

szymon-nagel-dt:
> I assume it's tested 😛 Can I check how it looks like? If you tried it on any tenant, the metrics with the dimension should appear, if I set the timeframe right.

Yes it now have 1 additional dimentsion for all monitored metrics. Origin. For most of them dimension is set to extension, autodiscovered have "autodiscovered dimension".

PR #485, opened 2024-02-08
PCLOUDS-3748: Adjusted align and reduce aggregations

joaquinfilipic-dynatrace:


heydarmd (approved):

szymon-nagel-dt (approved):

src/lib/metric_ingest.py
```diff
@@ -215,25 +213,50 @@ async def fetch_metric(
...
+    # Default, mainly for kind == DELTA (what we consider count)
+    reducer = 'REDUCE_SUM'
+
+    # Cannot be reduced
+    if metric_kind.lower().startswith('cumulative'):
+        reducer = 'REDUCE_NONE'
+    elif metric_kind.lower().startswith('gauge'):
+        reducer = 'REDUCE_COUNT_TRUE' if value_type.lower() == 'bool' else 'REDUCE_MEAN'
+
+    return reducer
```
emjot13-dynatrace, line 256: General comment about these 2 functions: would you consider creating dictionaries with the aligners and reducers? Then we will separate data from the functions, also you could assign keys like default_aligner, which often are better than comments. "Code never lies, comments sometimes do." 😄
Btw, if you do decide about using the dict, there is this [_MappingProxyType_](https://docs.python.org/3/library/types.html#types.MappingProxyType) which is immutable, which may be desired for such a use case.

src/lib/metric_ingest.py
```diff
@@ -350,7 +373,7 @@ def create_entity_id(service_name: str, service_dimensions: List[Dimension], tim
     return entity_id
 
 
-def convert_point_to_ingest_line(
+def _convert_point_to_ingest_line(
```
emjot13-dynatrace, line 376: Nice of you to include some code cleaning! 😄

src/lib/metric_ingest.py
joaquinfilipic-dynatrace: I agree with separating data from the functions, but I don't want to maintain keys that I won't use. As we have it now, I don't care for most of the possible values for the types, for instance. And I won't go half way, this is too simple for me to deserve it.

src/lib/metric_ingest.py
joaquinfilipic-dynatrace: We can do minor things like this in every PR, when we spot them

src/lib/metric_ingest.py
emjot13-dynatrace: Sure, valid points 😄

emjot13-dynatrace (approved):

PR #487, opened 2024-02-15
PCLOUDS-3555 - metrics filter out

nikita-paskal-dynatrace:


src/lib/configuration/config.py
```diff
@@ -42,6 +42,13 @@ def excluded_projects_by_prefix():
     return os.environ.get("EXCLUDED_PROJECTS_BY_PREFIX", "")
 
 
+def excluded_metrics_by_prefix():
+    return os.environ.get("EXCLUDED_METRICS_BY_PREFIX", None)
+
+def excluded_dimentions():
```
joaquinfilipic-dynatrace, line 48: Typo 😄 > dimension
Look for other occurrences.

k8s/helm-chart/dynatrace-gcp-monitor/values.yaml
```diff
@@ -132,6 +132,10 @@ scopingProjectSupportEnabled: "false"
 excludedProjects: ""
 # excludedProjectsByPrefix: comma separated list of projects substring that will be excluded from monitoring (e.g. "project-a,proj,pro").
 excludedProjectsByPrefix: ""
+# excludedMetricsByPrefix: comma separated list of metrics that will be excluded from monitoring (e.g. "function/execution_count,proxy/request_count,proxy/req").
+excludedMetricsByPrefix: ""
+# excludedDimensions: comma separated list of dimentions that will be excluded from monitoring in a format of "{gcp_metric}|{dimension}" (e.g. "loadbalancing.googleapis.com/l3/external/egress_bytes_count|resource.labels.client_country")
```
joaquinfilipic-dynatrace, line 137: I believe the example should be: "loadbalancing.googleapis.com/l3/external/egress_bytes_count|client_country"

joaquinfilipic-dynatrace: They shouldn't care about how to access the dimension, they only know the name of the metric and the name of the dimension, as it is stated in Google's docs.

heydarmd: Also it requires additional change in the code itself. [Link](https://github.com/dynatrace-oss/dynatrace-gcp-monitor/pull/487/files#diff-c5626096d0344873a917f85fe0608fc810b5b6e7f86b6474ca3929cc1b328dd3R195)

Because `dimension.key_for_fetch_metric` has the form of:

image


My guess is that we should split the dimension name by dot and get the last part of it
Probably using something like this:
```python
s[s.rfind('.') + 1:]
```

joaquinfilipic-dynatrace: So, the only thing I'm concerned about now is the distinction between metric and resource labels. The easy use case is taking care of filtering the metric labels. Can we make it for both? Does it make sense? Because filtering out instance_id means no entity, and if they think about that is because they have another problem. 
Let's go at least with metric labels.

joaquinfilipic-dynatrace: But maybe location... Not sure, if you can do both from code easily enough, let's contemplate both. We need to discuss, I think

PR #510, opened 2024-04-17
Bugfix - allow for using VPC networks with custom subnet creation

emjot13-dynatrace:
Add an option to use the VPCs created with custom subnets for cluster creation.
More context: when creating a VPC network the default option is creating custom subnets in this network,
then when this network is passed in the `values.yaml`, creating an autopilot cluster would fail because no subnet was specified.
Now it's possible to use VPC networks with both automatic and custom subnets.
Additionally, added some small cleaning, mainly breaking too long lines and fixing small inconsistencies.


szymon-nagel-dt (approved):
LGMT.
1 question: Any adjustment in docs required? if so it is also in scope of this bug?

joaquinfilipic-dynatrace (approved):

joaquinfilipic-dynatrace:
> LGMT. 1 question: Any adjustment in docs required? if so it is also in scope of this bug?

True, docs need the new params

PR #545, opened 2025-02-18
[DAQ-4166] Added option to include projects

Dawid-Kaszubski:
A copy of the community pull request: https://github.com/dynatrace-oss/dynatrace-gcp-monitor/pull/543
A Dynatrace developer responsible for this repository copied and pushed changes to allow the builds to pass. @ajth13 was consulted on the copy and creation of this PR. Thank you once again for your input in the dynatrace-gcp-monitor development. 

Original PR description:

> Where you can currently only exclude projects from monitoring this change adds the ability to specify a list to include.
> Business Justification
> 
> Enterprise customers with a large number of Dynatrace tenants and wants to be able to easily control which metrics are sent to which tenant and include list is much smaller than excluding all projects from other Dynatrace tenants.
> 
> Considerations
> 
> The project must still be included in the metric scoping this only filters the list of scoped projects.
> The exclude logic is preferred so if a project is in both lists it will not be included.

### DIFF
--- a/k8s/helm-chart/dynatrace-gcp-monitor/values.yaml
+++ b/k8s/helm-chart/dynatrace-gcp-monitor/values.yaml
@@ -175,6 +175,10 @@ scopingProjectSupportEnabled: "false"
 excludedProjects: ""
 # excludedProjectsByPrefix: comma separated list of projects substring that will be excluded from monitoring (e.g. "project-a,proj,pro").
 excludedProjectsByPrefix: ""
+# includedProjects: comma separated list of projects that will be included from monitoring (e.g. "project-a,project-b,project-c").
+includedProjects: ""
+# includedProjectsByPrefix: comma separated list of projects substring that will be included from monitoring (e.g. "project-a,proj,pro").
+includedProjectsByPrefix: ""
 # excludedMetricsAndDimensions: structure containing the metrics and dimensions that will be excluded from monitoring.
 # If you want to exclude a metric, add a "metric" entry. Any global metric matching its prefix will not be ingested.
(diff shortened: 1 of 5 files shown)


k8s/helm-chart/dynatrace-gcp-monitor/values.yaml
```diff
@@ -175,6 +175,10 @@ scopingProjectSupportEnabled: "false"
 excludedProjects: ""
 # excludedProjectsByPrefix: comma separated list of projects substring that will be excluded from monitoring (e.g. "project-a,proj,pro").
 excludedProjectsByPrefix: ""
+# includedProjects: comma separated list of projects that will be included from monitoring (e.g. "project-a,project-b,project-c").
+includedProjects: ""
```
joaquinfilipic-dynatrace, line 179: I would say there's no need for this, as long as we have "includedProjectsByPrefix". 
The thing is we have both for exclusion too, but that was because we firstly had "excludedProjects" and then some customer asked for excludedProjectsByPrefix". IMHO, we could stick only with the "...ByPrefix" ones.
But for the exclusion, removing one now means that some customers need to switch if they want to redeploy a new version.
To consider, at least.

ajth13: Yeah totally reasonable suggestion. I added both only to replicate the existing behavior (but inverted) but there is not really any need as you pointed out.

joaquinfilipic-dynatrace: Right. So, to me, it's fine either way. We can leave it as it is or clean it up a bit, with some note in the readme or somewhere.

joaquinfilipic-dynatrace (approved):

PR #577, opened 2026-01-12
DAQ-18531-gaps-in-gcp-metrics 

szymon-nagel-dt:
This pull request updates the initialization logic in `src/lib/context.py` to improve the handling of execution timestamps. The main change ensures that the `execution_time` is always rounded up to the next whole second, preventing potential gaps caused by timing drift.


src/lib/utilities.py
```diff
@@ -30,6 +30,24 @@ def chunks(full_list: List, chunk_size: int) -> List[List]:
     return [full_list[i:i + chunk_size] for i in range(0, len(full_list), chunk_size)]
 
 
+def percentile(sorted_data: list, p: float) -> float:
+    """Calculate percentile from sorted data using linear interpolation.
```
szymon-nagel-dt:
```suggestion
    """Calculate percentile from sorted data.
```

PR #622, opened 2026-07-03
fix: aggregate excluded cumulative dimensions

szymon-nagel-dt:
## Summary
- aggregate cumulative metrics with excluded dimensions so omitted labels are removed by GCP
- defensively skip excluded labels during Dynatrace dimension creation
- add unit coverage for cumulative dimension exclusion


kacperchojnacki-dnt (approved):
looks good!

PR #632, opened 2026-08-18
fix: do not drop resources missing a grouping label; add optional per-resource dt.security_context

davidg-datascene:
Three related changes to the `labelsGroupingByService` preview feature. The first is a bug fix and stands alone; the other two build on it and are both opt-in, defaulting to today's behaviour. Happy to split this into separate PRs if you would prefer to take them independently.

## 1. Resources missing a grouping label are silently dropped

When a service is listed in `labelsGroupingByService`, `fetch_metric` appends `aggregation.groupByFields=metadata.user_labels.<label>`. GCP omits any series whose resource has no value for that label, so those resources stop being ingested entirely — not one dimension missing, but every metric of that service for that resource.

Measured against a real project: 31 series ungrouped, 29 grouped. The two absent resources were simply unlabelled.

`src/main.py` appends `NO_GROUPING_CATEGORY` only in the `if not groupings:` branch, so a listed service never gets an ungrouped pass to fall back on.

**Fix:** run the grouped fetch, record which resources it returned, then repeat the query without the group-by and emit only the resources the grouped pass never returned. Unlabelled resources keep being ingested; they simply carry no user-label dimension.

Dedup is keyed on `(resource.type, sorted(resource.labels))` and is required rather than cosmetic — grouped and ungrouped rows for the same resource differ by one dimension, so without it every labelled resource would ingest as two distinct series.

Cost is one extra `timeSeries.list` call per **listed** service. Services not listed take the same single-pass path as before, byte for byte.

## 2. `DT_SECURITY_CONTEXT_USER_LABEL`

`dt.security_context` is currently a per-deployment constant applied to every datapoint. This adds an optional env var naming a GCP user label whose value is used instead, per resource, falling back to `DT_SECURITY_CONTEXT` when the resource has no such label.

The motivation is a single project containing resources owned by different teams: the metrics can then carry the security context of the resource's owner rather than of whoever deployed the monitor.

Unset by default, so behaviour is unchanged unless explicitly enabled. It depends on change 1 — without the backfill, enabling it would drop every unlabelled resource.

### Worked example

One project, two teams owning the same resource type, and a third resource nobody has labelled:

| Cloud SQL instance | GCP user label |
| --- | --- |
| `db-alpha` | `eai: 12345` |
| `db-beta` | `eai: 7801234` |
| `db-gamma` | *(none)* |

```yaml
dtSecurityContext: "1000000"        # deployment-wide fallback, as today
dtSecurityContextUserLabel: "eai"   # per-resource override
groupAllServicesByUserLabel: "eai"  # make the label available on every service
```

Two calls result:

```
pass 1  aggregation.groupByFields=metadata.user_labels.eai
        -> db-alpha, db-beta          (db-gamma omitted by GCP: no value for the label)

pass 2  no groupByFields                                        [backfill, change 1]
        -> db-alpha, db-beta, db-gamma
           db-alpha/db-beta skipped, already seen in pass 1
           db-gamma emitted
```

Ingested result:

| Instance | `dt.security_context` | `eai` dimension | Source |
| --- | --- | --- | --- |
| `db-alpha` | `12345` | `12345` | its own label |
| `db-beta` | `7801234` | `7801234` | its own label |
| `db-gamma` | `1000000` | *(absent)* | `dtSecurityContext` fallback |

Each team's metrics carry that team's context, and the unlabelled resource keeps flowing rather than disappearing. Adoption is therefore incremental: nothing has to be labelled before the feature is switched on, and each resource moves to its own context as it gets labelled.

Note that `dtSecurityContextUserLabel` does nothing on its own — GCP only populates `TimeSeries.metadata.userLabels` for labels named in the aggregation, so the label must also be grouped on, by either `groupAllServicesByUserLabel` or a per-service `labelsGroupingByService` entry. Called out in `values.yaml`.

## 3. `GROUP_ALL_SERVICES_BY_USER_LABEL`

Groups every service by a given user label, so a deployment wanting one label across the board does not have to enumerate every service in `labelsGroupingByService`. Services listed there keep their own grouping, so the global acts as a default rather than an override.

Unset by default. Enabling it puts every service on the two-pass path, so `timeSeries.list` volume roughly doubles — noted in `values.yaml`.

## 4. One grouping per service

Each entry under a service's `groupings` is queried separately, so declaring several fetches and ingests every matching resource once per grouping. `values.yaml` warned about this, but nothing enforced it, and change 1 makes it worse: dedup lives inside a single `fetch_metric` call and does not span groupings, so a resource carrying none of the labels would be emitted by every ungrouped pass.

`read_labels_grouping_by_service_yaml` now merges a service's groupings into one comma-separated grouping, keeping every requested label, deduplicating repeats and logging the merge.

## Tests

`tests/unit/test_security_context_user_label.py` (new) covers the backfill, the security context and the fallback chain; `tests/unit/test_labels_grouping_by_service.py` is extended for the global label and the grouping collapse.

145 unit tests pass. Note `tests/unit/extraction_rules` fails to collect with `ModuleNotFoundError: No module named 'unit'` both on this branch and on unmodified `master`, so it looks pre-existing rather than related to these changes. The suite needs `PYTHONPATH=src`.

Chart rendering verified with defaults, with the new values set, and with `deploymentType=logs` — the new ConfigMap keys are metrics-only, so the last confirms the logs deployment gets no dangling `configMapKeyRef`.

## Compatibility

Both new variables default to empty. With them unset the only behavioural changes are the backfill fix and the grouping collapse, both of which affect only services already listed in `labelsGroupingByService`. The `DT_SECURITY_CONTEXT` → project id fallback is unchanged.

### DIFF
--- a/k8s/helm-chart/dynatrace-gcp-monitor/templates/function-config.yaml
+++ b/k8s/helm-chart/dynatrace-gcp-monitor/templates/function-config.yaml
@@ -34,6 +34,8 @@ data:
   INCLUDED_PROJECTS_BY_PREFIX: {{ .Values.includedProjectsByPrefix | quote }}
   KEEP_REFRESHING_EXTENSIONS_CONFIG: {{ .Values.keepRefreshingExtensionsConfig | quote }}
   {{- if or (eq .Values.deploymentType "metrics") (eq .Values.deploymentType "all") }}
+  DT_SECURITY_CONTEXT_USER_LABEL: {{ .Values.dtSecurityContextUserLabel | quote }}
+  GROUP_ALL_SERVICES_BY_USER_LABEL: {{ .Values.groupAllServicesByUserLabel | quote }}
   PRINT_METRIC_INGEST_INPUT: {{ .Values.printMetricIngestInput | quote }}
   METRIC_INGEST_CONCURRENT_PUSHES: {{ .Values.metricIngestConcurrentPushes | quote }}
   METRIC_INGEST_BATCH_SIZE: {{ .Values.metricIngestBatchSize | quote }}
--- a/k8s/helm-chart/dynatrace-gcp-monitor/templates/function-deployment.yml
+++ b/k8s/helm-chart/dynatrace-gcp-monitor/templates/function-deployment.yml
@@ -207,6 +207,16 @@ spec:
             configMapKeyRef:
               name: dynatrace-gcp-monitor-config
               key: DT_SECURITY_CONTEXT
+        - name: DT_SECURITY_CONTEXT_USER_LABEL
+          valueFrom:
+            configMapKeyRef:
+              name: dynatrace-gcp-monitor-config
+              key: DT_SECURITY_CONTEXT_USER_LABEL
+        - name: GROUP_ALL_SERVICES_BY_USER_LABEL
+          valueFrom:
+            configMapKeyRef:
+              name: dynatrace-gcp-monitor-config
+              key: GROUP_ALL_SERVICES_BY_USER_LABEL
         volumeMounts:
         - mountPath: /code/config/activation
           readOnly: true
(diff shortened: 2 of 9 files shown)


davidg-datascene (commented):
@szymon-nagel-dt - Hi, any progress with the review? Thanks

feespina2 (approved):

davidg-datascene:
Closing PR as replaced by - https://github.com/dynatrace-oss/dynatrace-gcp-monitor/pull/637

PR #637, opened 2026-09-02
Per-resource dt.security_context from a GCP user label, opt-in backfill of unlabelled resources

szymon-nagel-dt:
This PR keeps the per-resource security context and missing-label backfill from #632, with both features disabled by default. Existing deployments do not need to change their configuration.

## Changes

- `dtSecurityContextUserLabel` (default: `""`) uses a GCP resource label as the metric's `dt.security_context`. For example, setting it to `owner` maps `owner=team-a` to `dt.security_context=team-a`. The monitor requests the label automatically and backfills resources missing the grouping labels with the default context.
- `includeResourcesWithoutGroupingLabels` (default: `false`) enables backfill for existing `labelsGroupingByService` configurations, without enabling label-based security contexts.
- If none of a metric's grouped queries returns data but backfill does, the monitor logs a message once per project and grouping combination. This can help identify a misspelled label name.

Each metric now has one shared backfill query, after all its grouped queries and pages finish. A resource returned by any grouping is excluded from backfill, so a resource missing every grouping label is no longer backfilled separately for each grouping. If a grouping fails, successful results are kept but backfill is skipped rather than assigning potentially labelled resources to the default context. A failed backfill also leaves successful grouped results intact.

The global grouping option and automatic merging of multiple groupings from #632 are not included. Existing groupings retain their separate queries and results.

## Before enabling

Backfill adds one ungrouped query per affected gauge/delta metric, including any pagination. For a single grouping this can roughly double Monitoring API requests. Backfill can also increase the number of metrics ingested into Dynatrace.

Label-based security contexts change metric series identity and can affect access under your Dynatrace permission policies. Configure permissions first and check who can edit resource labels in GCP. Cumulative metrics still use the default security context and do not make the extra query. Usage examples and warnings are in `values.yaml`.

## Verification

- 176 unit tests passed locally, covering multiple groupings through the real scheduler, shared backfill, pagination, failures, cancellation, disabled defaults, and cumulative aggregation.
- The local metrics integration test hit WireMock response timeouts. The same failure was reproduced on the previous PR commit (`d2e34b2`), before this fix.
- Helm templates were previously verified for `logs`, `metrics`, and `all`; this fix does not change the chart.
- Earlier read-only REST checks confirmed missing-label filtering on GCE and Pub/Sub, metadata retrieval through grouping, and ignored grouping under `REDUCE_NONE`. Two labels in one grouping required both; a nonexistent label returned no series and no error.

## Limitation

Backfill identifies resources by their returned resource type and labels. If aggregation removes the labels needed to distinguish resources, it may skip unlabelled resources that share a key with a labelled result. The live checks covered GCE and Pub/Sub, not every supported resource type.


src/lib/metric_ingest.py
```diff
@@ -396,59 +401,120 @@ async def fetch_metric(
...
+                        else:
+                            lines.append(line)
+
+            next_page_token = page.get('nextPageToken', None)
+            if next_page_token:
+                _update_params(next_page_token, page_params)
+            else:
+                should_fetch = False
+
+    seen_resource_keys = set()
```
kacperchojnacki-dnt, line 538: A resource with no grouping labels at all would get backfilled separately by every grouping, right? Looks like double ingestion

kacperchojnacki-dnt (approved):

PR #638, opened 2026-09-17
DAQ-29625: scope autodiscovered-metric filter inheritance to the linked service's own metrics

kacperchojnacki-dnt:
## Summary
- `fetch_metric` (`src/lib/metric_ingest.py`) no longer applies an extension service's `filter_conditions` to autodiscovered metrics outside that service's own metric family — scoped via a new `_resolve_autodiscovered_metric_filter` helper that matches on GCP API prefix (`google_metric` up to the first `/`). Fixes user-defined log-based metrics (`logging.googleapis.com/user/*`) being silently filtered out by an unrelated extension's `(resource.labels.env=...)` condition.
- Same-API autodiscovered metrics (e.g. `apigee.googleapis.com/proxy/details`, DAQ-21921) keep inheriting the filter — this is not a revert of PR #596.
- Added a log line when an inherited filter is applied, and a `WARNING` log when a metric query returns zero `timeSeries` while any filter (inherited or native) is active — this failure mode previously left no trace, which is what made the underlying RFA (DAQ-28706) take four weeks to diagnose.

## Safety
- Change is scoped to `src/lib/metric_ingest.py`'s autodiscovery filter-selection branch only; the non-autodiscovery path (`service.monitoring_filter`) is untouched, verified by `test_non_autodiscovered_metric_still_uses_own_service_filter`.
- DAQ-21921 non-regression is covered by an explicit regression test (`test_daq_21921_regression_same_api_autodiscovered_metric_still_inherits_filter`).
- No other repo consumes this code (standalone Python service). Full suite: 156 passed, 0 failed.

## Where to review
Start at `src/lib/metric_ingest.py`: `_gcp_api_prefix`/`_resolve_autodiscovered_metric_filter` (new helper) and their call site in `fetch_metric`. Key invariant: filter inheritance is now gated on API-prefix match against the linked service's *own* metrics, not on "any metric sharing the monitored resource type" as before. `tests/unit/test_autodiscovery_filter_inheritance.py` covers both the fix and the two non-regression cases (DAQ-21921, native-filter path).


src/lib/metric_ingest.py
```diff
@@ -90,6 +90,41 @@ def should_exclude_dimension(
...
+    for a different metric family (e.g. a user-defined logging.googleapis.com/user/* metric linked
+    to an apigee.googleapis.com/* extension service) and querying with it silently drops all data.
+    """
+    if not linked or not linked.possible_service_linking:
+        return ""
+
+    linked_service = linked.possible_service_linking[0]
+    linked_service_prefixes = {_gcp_api_prefix(m.google_metric) for m in linked_service.metrics}
+    if _gcp_api_prefix(metric.google_metric) not in linked_service_prefixes:
+        return ""
```
szymon-nagel-dt: Could we check the API prefixes across the linked feature sets belonging to the same service, rather than just the first entry? Each `GCPService` represents one service/feature-set combination. For example, `gce_instance/default_metrics` contains `compute.googleapis.com/*`, while `gce_instance/agent` contains `agent.googleapis.com/*`.

With `default_metrics` first, an autodiscovered agent metric loses the configured resource filter even though the agent feature set is enabled. I reproduced this: reversing the candidates changes whether the zone filter is retained.

Could we include the sibling feature sets in the prefix check and add a test covering both orders? No need to change how the filter itself is selected or combine unrelated services.

szymon-nagel-dt (approved):
LGTM

